Issue #70, opened 2021-01-21
support keyword sets in record model


tomkralidis:
Currently in the record model, we have a model which results in the following for keywords:

```
"properties": {
    ...
    "keywords": ["foo", "bar", "baz"],
    "keywords-codespace": "https://codes.example.org/vocab1"
    ...
 }
```

It is common for metadata providers to provide "sets" of keywords. For example, in MSC metadata we provide the following sets of keywords:
- open data freetext type keywords (Canadian core subject thesaurus)
- WMO vocabulary keywords
- etc.

ISO 19115 allows for 1..n sets of keywords.  The value proposition here is that metadata providers can manage their metadata once and provide keyword sets specific to a given community, thus publishing to multiple catalogues' keyword requirements.

In our MetOcean Best Practices for OGC API - Records work, we have also [identified this as key](https://github.com/OGCMetOceanDWG/ogcapi-records-metocean-bp/issues/4#issuecomment-764756301) for a given metadata record.

```
  "properties": {
    ...
    "keywords": [
      {
        "keywords": [ "foo1", "bar1", "baz1"],
        "codespace": "https://codes.example.org/vocab1"
      },
      {
        "keywords": [ "foo2", "bar2", "baz2"],
        "codespace": "https://codes.example.org/vocab2"
      }
    ]
    ...
  }
```

For an `items` query, a client querying `keywords` would be querying across all the given sets.

Is this something we should consider in core or is it asking too much?

We could certainly have this as an extension in our best practice (i.e. `metoc:keywords`), but it may provide value to other information communities, hence this proposal.  For simple requirements, the above would simply be a single set/object.

Thoughts?  I'm happy to update the specification/schemas/etc. here if there is consensus that this would be valuable in core.

m-mohr:
In STAC we just have a field with keywords, which I think is enough for the core. Codespaces/vocabulary sounds like an extension to me.

chris-little:
As there are a number of overlapping authoritative vocabularies used in real-world safety critical regimes, I suggest it is in core, but as a separate conformance class. 
Example 1: meteorology&aviation&volcanoes.
Example 2: meteorology&nucleotides&chemicals

pvgenuchten:
[...]

PR #84, opened 2021-02-05
resolve extents to OGC API - Common, Collections, make optional

tomkralidis:
Updated extents schema reference.  I also made `extents` optional, seeing that the extents schema itself has no required properties, which may end up allowing things like:

```yaml
extents:
```

...just so documents can pass schema validation without added value?


cportele:
+1 to making `extent` optional. If there is no extent there is no point in having an empty object.

tomkralidis:
UPDATE 2021-02-09: since rebasing from @pvretano 's recent updates, this PR now updates the record model and example per #70, and adds a feature collection schema for Records responses.

PR #89, opened 2021-02-15
use id for collection in example openapi document

### LINKED ISSUE #88: collection id or name
The example yaml uses name to identify a collection
https://github.com/opengeospatial/ogcapi-records/blob/076674b82323ab6971df4f333c06a900a9af395a/core/openapi/openapi.yaml#L328-L367
but the spec indicates to use 'id' instead
https://docs.opengeospatial.org/DRAFTS/20-004.html#collections-metadata

pvgenuchten:
fix #88

### DIFF
--- a/core/openapi/openapi.json
+++ b/core/openapi/openapi.json
@@ -704,10 +704,10 @@
         }
       },
       "collectionInfo" : {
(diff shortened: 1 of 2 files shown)


pvgenuchten:
fixed already

PR #97, opened 2021-02-22
quote pattern in schema

tomkralidis:


pvretano:
Tom, the sortBy pattern is actually not correct (typo on my part) so I'll merge and then fix the pattern in the master branch.

Issue #52, opened 2020-10-01
include facet stats in a search result


pvgenuchten:
This is probably relevant for oapi-common, but i want to raise it here first, because the faceted search concept is currently most common in the catalogue domain, but i can see similar uses in oapif & sensor api's. 

A typical use case in catalogues is [faceted search](https://en.wikipedia.org/wiki/Faceted_search) to further limit a search result. Faceted search requires an analyses of the search result best managed during the backend query, resulting in some statistics being added to a search result. 

What would be the best approach/location in a typical OAPIR search response to include this information and does it make sense to suggest this as a common extension to OAPI?

mhogeweg:
I see facet information as a property of the catalog as a whole (or of a collection), not so much specific to a search. when performing a search, selections made in facets translate to a search expression that could also have been generated based on a different UI or based on CQL/Filter style queries.

pvgenuchten:
my request relates to facet information specific to the search result, so you query for "forest", then the search result will return a first page of results plus facet statistics on the total resultset (how many maps/services/datamodels/datasets, how many results in projection a/b/c, how many results in language eng/fre/ger, etc). See for example the geoss portal
![image](https://user-images.githubusercontent.com/299829/94899780-2f4a4700-0494-11eb-9b05-dea57b2d3898.png)
for relevant protocols/keywords is per search result indicated how many results in the current result set

mhogeweg:
ok, I think that would make for a nice extension. It would require some discussion as not all facets are as clean as a term+count. In Geoportal Server we support temporal, numeric, and spatial facets (see http://geoss.esri.com/geoportal_264):

![image](https://user-images.githubusercontent.com/394890/94955397-59553680-049f-11eb-94d5-183f6ac99ae1.png)

We see faceted search as a 180 degree turn from the traditional 'google one box' search experience. Rather than guessing terms to look for something, you start with all the catalog has to offer. Through the choices made in the facets, you discard what is not relevant, whether it be spatially outside your area of interest, or temporally not the right period, or not the right service type, etc etc. Every choice made in one facet also filters content in the other facets, always leaving you with what matches all of the choices. You can quickly broaden the view by removing a facet choice.

Different catalogs (collections?) may include different facets, suitable for the content in the catalog. for example, same technology, now showing information from a series of Spatio-Temporal Asset Catalogs: http://geoss.esri.com/imagecatalog/

[...]

PR #110, opened 2021-04-10
extension to facilitate faceted search

pvgenuchten:
Initial work on #52

Current implementation uses terminology inspired by Elastic API. Since it is a common catalogue backend. 

The work may also be relevant for ogc-api-common (ogc-api-features, ...). However faceted search is currently more common in catalogues.

### DIFF
--- a/proposals/aggregations/README.md
+++ b/proposals/aggregations/README.md
@@ -1,0 +1,135 @@
+# OGC API - Records - Term Aggregations
+
+This folder contains the content for the standard extension OGC API - Records - Term Aggregations.
+
+# Overview
+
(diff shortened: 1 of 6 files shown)


pvretano:
Paul, would it be too much trouble to move this to the proposals directory?  I want to keep the extensions directory reserved for extensions that the SWG is tasked to work on in the charter like the collections extension.  Extensions like faceted search are proposed extensions and we will need to revise the charter before that SWG can "officially" work on them.  Once we get close to completing core and the collections extension, we can discuss additional extensions such as the ISO 19XXX extension, transactions, harvesting and of course faceted search.
If merging the PR first would make the move easier then go ahead and do that.

pvgenuchten:
moved to proposals

PR #133, opened 2021-08-09
update type and license

### LINKED ISSUE #99: Re-consider all queryables that are links ...
@cportele mentioned that there are several queryables (license, publisher, etc.) that are links that could, in fact, be included as associations in the associations sections with the appropriate rel.  It is a bit of a grey area but we should revisit these and see if this is something we might want to consider to make the core ever more lean.

cportele:
closes #99

In the issue we recorded that the links should be in "links", so I have implemented this, but shouldn't the type and license links be in "associations"?

### DIFF
--- a/core/openapi/schemas/recordGeoJSON.yaml
+++ b/core/openapi/schemas/recordGeoJSON.yaml
@@ -35,8 +35,11 @@ properties:
       type:
         type: string
         description: |-
-          The nature or genre of the resource.
-        format: uri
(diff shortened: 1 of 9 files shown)


tomkralidis (commented):
Thanks @cportele, looks good to me. My only comment is whether we should cite/reference SPDX?

cportele:
Good point @tomkralidis - I added links to SPDX and also added SPDX to the bibliography

PR #136, opened 2021-08-27
camel case, and example updates

tomkralidis:


tomkralidis:
@pvretano let me know if this PR needs any adjustments, thanks.

Issue #137, opened 2021-08-27
consider qualification of externalId


tomkralidis:
In the [core record model](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/recordGeoJSON.yaml#L74-L80), `externalId` is defined as an array.  Should we consider qualifying lists of `externalId`'s to be able to delineate accordingly?  For example, for WMO station metadata, we have a number of identifiers (WIGOS ID, ICAO ID, etc.), i.e.:

```json
"externalId": [
    {
        "scheme": "WIS",
        "value": "urn:x-wmo:md:int.wmo.wis::ca.gc.ec.msc-1.1.1.1.10 "
    },
    {
        "scheme": "ICAO",
        "value": "foo123"
    }

]
```

pvretano:
13-SEP-2021: @tomkralidis to make proposal for consideration

tomkralidis:
Proposal in #142

tomkralidis:
Update from SWG meeting 2021-11-15: agreed/consensus to merge, via #143

PR #142, opened 2021-09-13
qualify record model externalId (#137)

tomkralidis:
Addresses #137

### DIFF
--- a/core/examples/json/record.json
+++ b/core/examples/json/record.json
@@ -21,7 +21,12 @@
             "saoz"
         ],
         "language": "en",
-        "externalId": "urn:x-wmo:md:int.wmo.wis::https://geo.woudc.org/def/data/ozone/total-column-ozone/totalozone",
+        "externalId": [
+            {
+                "scheme": "WMO:WIS",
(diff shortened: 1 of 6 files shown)


pvretano (approved):

cportele:
@tomkralidis - The externalId query parameter values are compared against the "value" member and "scheme" is ignored, correct?

pvretano:
@cportele Yes, and I will add that clarification to the document.

pvretano:
@tomkralidis the descriptive text for the value of the external identifier reads:
```
An identifier for the knowledge organization system used to classify the resource.  It is recommended that the
identifier be a resolvable URI
```
I would reword this slightly to not use the term "classify the resource".  That terminology bumps into the purpose of the `themes` core queryable.  Rather, the text should say something like:
```
A reference to an authority or identifier for a knowledge organization system from which the external identifier was obtained.  It is recommended that the identifier be a resolvable URI.
```
If you agree, can you update the text directly in the master branch?  Sorry about that, I should have caught that before merging the PR.

tomkralidis:
@pvretano done in #143

PR #143, opened 2021-09-13
update record model externalId.scheme (#137)

tomkralidis:
Addresses https://github.com/opengeospatial/ogcapi-records/pull/142#issuecomment-918242295


tomkralidis:

Issue #26, opened 2020-02-25
Define standard identifiers for OGC (and other) resources.


pvretano:
This is a branch of issue #13 .  

There is a need to define standard identifiers for OGC resources and perhaps other resources too.  Some work has already been done in this area both inside and outside the OGC.  The following source material is a starting point:

- https://portal.opengeospatial.org/files/?artifact_id=58896&version=1
- http://docs.opengeospatial.org/per/16-043.html
- https://github.com/OSGeo/Cat-Interop/blob/master/LinkPropertyLookupTable.csv

During the 24-FEB-2020 teleconference the issue of governance arose.  This list of standard identifiers should be easily maintainable.  The suggestion was made that a document maintained on github might be a good approach.  There was also a discussion about the fact that the OGC Naming authority would need to be involved.

mhogeweg:
easily maintained would not exclude good governance. who 'owns' this list and approves additions/modifications?

dr-shorthair:
This what OGC-NA does - @rob-metalinkage @ghobona

pvretano:
@mhogeweg Dunno. We need to speak with the OGC Naming Authority to see (a) if they would be the "owners" of this list (makes sense to me) and (b) how they would propose to manage it.  As mentioned, there have been several attempts both in and of the OGC in this direction and in today's SWG discussion it was agreed that perhaps we can push the effort into a more formal posture addressing the ownership and maintenance issues.  Like many current efforts in the OGC API domain (e.g. CQL, etc.) we can start the work of getting the list defined in this SWG and then hand it off to the appropriate owner for ongoing maintenance.  I have already contacted OGC to give them a heads up about this issue.

pvretano:
@dr-shorthair Actually I contacted Gobe and Scott.  I'll forward the message to Rob as well.

rob-metalinkage:
hi all,

as it happens I am aiming to take an analysis of the current heterogeneity to a side-meeting at the OGC staff retreat in March. This is good timing to get community pull on the thread!!!

I am keen to do multiple things here:
1) Find all the legacy identifiers in use and cross-reference them in the definitions server (a semantic resource) (sameAs relationships) - and identify canonical URIs for each resource type (if possible)
2) Enrich the model by declaring other relationships between specifications (profileOf, supercedes, uses, related to) and links between docs and Conformance Classes
[...]

PR #145, opened 2021-09-19
add Annex on resource types (#26)

tomkralidis:
Addresses #26.  Note that there are pointers to codelists (vs. listing out all the actual types as examples).  I went this way given it would be confusing to have numerous examples of how to say a link is a WMS, for example, from the various codelists.


tomkralidis:

PR #146, opened 2021-09-23
Example for dataset record/collection

cholmes:
Using a geopackage, with data for LINZ.

Would be great to have a review, as I was trying to adjust lots of little details to get right, and I'm sure I missed some.


core/examples/json/crawlable_catalogue/dataset_example/nz-buildings-collection.json
```diff
@@ -0,0 +1,72 @@
...
+    "links":
+    [
+        {
+            "href": "https://www.linz.govt.nz/data/licensing-and-using-data/attributing-linz-data",
+            "rel": "license",
+            "type": "text/html",
+            "title": "license"
+        },
+        {
+            "href": "buildings-record.json",
```
cportele, line 53: Should be `"href": "nz-buildings-record.json",`

core/examples/json/crawlable_catalogue/collection.json
```diff
@@ -0,0 +1,32 @@
...
+    {
+        "spatial":
+        {
+            "bbox":
+            [
+                [
+                    -180,
+                    -90,
+                    180,
+                    90
```
cportele, line 15: This should be the minimal extent that includes all records. Or will the imagery collection cover the whole world?

core/examples/json/crawlable_catalogue/dataset_example/nz-buildings-record.json
```diff
@@ -0,0 +1,93 @@
...
+            {
+                "name": "producer"
+            }
+        },
+        "formats": [
+            "GeoPackage",
+            "GeoJSONSeq",
+            "FlatGeobuf"
+        ],
+        "license": "CC-BY-4.0",
```
cportele, line 64: Remove the comma and close the "properties" object (add "}").

core/examples/json/crawlable_catalogue/collection.json
cholmes: The imagery one is marked as the whole world, so since this was a super class it seemed like it should include it all. I agree a better example would be a limited imagery catalog with valid bounds.

cholmes:
Ok, finally got to fixing these, sorry for the delay. I think it's ready to go. Didn't change the world bounds though, as I wasn't sure if the imagery directory is going to get more examples. @pvretano - are you going to add more? If not I can clean it up (remove all the collections that don't link anywhere) and set the proper bounds throughout.

Issue #122, opened 2021-06-21
JSON/YAML schema null support


tomkralidis:
Many of the JSON/YAML schema properties have a single type ([example](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/recordGeoJSON.yaml#L40-L43)).

When doing validation testing against a sample instance document with, say, `"title": null`, the JSON schema validator throws error.  If I update `core/openapi/schemas/recordGeoJSON.yaml` (`properties.title`) to allow nulls, i.e.:

```yaml
title:
  type:
    - string
    - null
  description: |-
    A human-readable name given to the resource.
```

The JSON instance document then validates fine.

What is the general guidance on nullable properties?

cc @pvretano @cportele

pvretano:
@tomkralidis I am not sure what the general guidance is BUT I can say that I don't think we explicitly specify that a property can be null in features (e.g. see [collection.yaml](http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/collection.yaml)).  The same is true for processes too.

cportele:
@tomkralidis - In general, we have only used null as a type where this is syntactically required by the underlying spec (e.g., GeoJSON) or where there is a semantic difference between "not present" and "null" (e.g., PATCH with JSON Merge Patch). In all other cases the preferences is to simply drop a property, if it has no value.

Also note that, as @pvretano has pointed out, OpenAPI 3.0 schemas do not support null as a type (it uses a nullable property instead), this is different from JSON Schema. This difference wasn't the driver for the approach taken, but it helps so that many of our OpenAPI schemas are also valid JSON schemas.

pvretano:
04-OCT-2021: @tomkralidis will add some text to the specification about null handling.

tomkralidis:
PR in #148

pvretano:
07-FEB-2022:  @tomkralidis PR #148 resolves this issue.

PR #148, opened 2021-10-04
add note about JSON null values (#122)

tomkralidis:
As per #122


tomkralidis:
Thanks @cportele.  I've updated this PR to this effect, for review/comment.

Issue #147, opened 2021-10-03
consider temporal resolution


tomkralidis:
For metadata records which are not split temporally by some resolution, should we consider temporal resolution (a la `dcat:temporalResolution`)?  Imagine for data which updates daily/hourly/etc.:

```json
                "temporal": {
                    "interval": [
                        [
                            "2000-10-30T18:24:39+00:00",
                            "2007-10-30T08:57:29+00:00"
                        ]
                    ],
                    "resolution": "P1D"
                }
```

pvretano:
04-OCT-2021: @tomkralidis to write up a PR to consider.

tomkralidis:
PR in #149

tomkralidis:
Update from SWG meeting 2021-11-15: agreed/consensus to merge (done via #149)

PR #149, opened 2021-10-04
add extent.temporal.resolution property (#147)

tomkralidis:
As per #147


tomkralidis:

PR #179, opened 2022-09-14
Issue 178 - Move 'created' and 'updated' fields from the record to the links pointing to the representation of the record.

pvretano:
Implement change discussed at the Sept. 2022 OGC Metadata Code Sprint.


core/examples/json/record.json
```diff
@@ -32,8 +32,8 @@
     ]
   },
   "properties": {
-    "recordCreated": "2021-02-08",
-    "recordUpdated": "2021-02-08",
+    "created": "2021-02-08",
```
m-mohr, line 35: All the values for created and updated in the examples are invalid, I think.
The spec says they are "datetime"s, but the examples are "date"s only. So I think it should be something like "2021-02-08T00:00:00Z"

m-mohr:
LGTM except that the dates in the examples should be valid datetimes.

m-mohr:
LGTM, thanks!

Issue #182, opened 2022-09-28
Definition of a Parent Identifier/Collection core Record property


kalxas:
In pycsw we are implementing the relation between a collection and a record using the MD_Metadata:ParentIdentifier element from ISO-19115.
While implementing STAC Item support, we realized that we do not have a core element in OGC API Records to describe the parent identifier or the collection the record belongs to. Internally we are still using the ParentIdentifier but without a record element, such a relationship would not be defined in a transaction.
Looking at STAC we can find this in the Item properties, named as "collection":
https://github.com/radiantearth/stac-spec/blob/master/examples/simple-item.json#L42
https://github.com/radiantearth/stac-spec/blob/master/examples/extended-item.json#L83

I am wondering if we should also define this "collection" as a Record core element and if this would also help the harmonization effort #178

cportele:
OGC API Features/Records uses the IANA "collection" link relation to point to the collection that the item (record, feature, ...) belongs to. STAC does that, too, so I do not think there is harmonization issue here. 

In addition, STAC adds the (redundant) "collection" member. Such a member could be added in Records, too, if there is a need, but it should be clear why the redundancy is necessary for Records. What is the main use case where the link is insufficient?

kalxas:
I agree that his is a redundant element if the link exists. Perhaps it has some value as an optional element for mapping to STAC "collection" and ISO-19115 "parentidentifier". Should we leave this to the implementation side and not mention it in the spec?

cportele:
My view is that if the additional member has value in typical usage scenarios, then it probably should not be optional as otherwise clients cannot rely on it. If not, we should not add/discuss it.

cportele:
That would be invalid in OGC API Features (the collection link is required) as well as in STAC (the member must be present, if the link is there, and otherwise the member is not allowed).

pvretano:
14-NOV-2022:  Recommendation 4 in the Part 1 should be a requirement.  @pvretano will fix.

kalxas:
Nice to discuss this again in Common SWG meeting :)

PR #183, opened 2022-09-28
Adding collection property

kalxas:


kalxas:
Implements #182

pvretano:
@kalxas apologies if this has already been condiered but why can't we just have a link in the links section with rel=collection?

PR #185, opened 2022-10-05
Declaring a default sort order.

pvretano:
This PR add support for a server that support sorting to declare the default sort order.


core/examples/json/defaultSortOrder.json
```diff
@@ -0,0 +1,14 @@
+{
+    "id": "sentinel1cat",
+    "itemType": "record",
+    "title": "Sentinel-1 Catalogue",
+    "description": "A sample catalogue of Sentinel-1 products stored in S3 on AWS.",
+    "defaultSortOrder": [ "updated-", "area-"],
```
cportele, line 6: The sortby parameter prepends the order indicator. To be consistent this should be `[ "-updated", "-area"]`.

pvretano: Doh!  You are right ... I fix!

pvretano:
14-NOV-2022: SWG agreeses this should be merged keeping in mind that it will eventually end up in Features.

Issue #188, opened 2022-11-25
extend themes/concepts


tomkralidis:
Further to http://docs.ogc.org/DRAFTS/20-004.html#sc_keywords_and_themes, Recommendations 1A and 1B state:

<html><body>

Recommendation 1 | /rec/core/record-associations
-- | --
A | Implementations SHOULD use the record properties.keywords property to provide free-form keywords associated with the current record.
B | Implementations SHOULD use the record properties.themes  property to enumerate 1..n objects of concepts and their respective  knowledge organization system/controlled vocabulary associated with the  current record.

</body>
</html>

For 1B, should we consider providing more information for a given concept to help further qualify same?  For example:

```json
{
  "themes": [
    {
      "concepts": [
        {
          "name": "OCEANS",
          "url": "https://gcmd.earthdata.nasa.gov/kms/concept/91697b7d-8f2b-4954-850e-61d5f61c867d"
        }
      ],
      "scheme": "https://gcmd.earthdata.nasa.gov/kms/concepts/concept_scheme/sciencekeywords"
    }
  ]
}
```

The idea here is that if a user is going to use themes/concepts (vs. freetext keywords), it would make sense to extend per above, and cover use cases where a dataset's underlying attributes/parameters/variables could be provided with deeper meaning.  Note that this approach is also put forth as part of ISO 19139 (`gmx:Anchor` elements).

In summary, adding a `scheme` in keywords is valuable for referring to the associated knowledge organization system.  Adding qualifiers to `concepts` would provide a robust, generic and powerful approach in support of semantics.

Happy to put together an associated PR.

pvretano:
@tomkralidis sounds good to me but lets discuss at the next meeting and see what the rest of the SWG thinks.  If everyone is good with it, you can then go to the trouble of creating the PR.

kalxas:
I also support this suggestion.
+1
Angelos

cportele:
An alternative representation for a keyword that has a URI would be a link with link relation type "tag". But I guess this is not an option, because I assume we still want to allow keywords in a scheme, but without a URI.

pvretano:
28-NOV-2022: SWG merges PR. @tomkralidis please review and close the issue.

PR #190, opened 2022-11-27
extent themes/concepts (#188)

tomkralidis:
Addresses #188


pvretano:
28-NOV-2022: SWG agrees this is a good change.  Merging.

PR #193, opened 2022-12-03
fix properties.externalIds definition

tomkralidis:
Fixes schema definition for `properties.externalIds` (`value` should be a sibling of `scheme`).


pvretano:
12-DEC-2022: Just fixes indentations.

PR #197, opened 2022-12-19
Add organization/affiliation property in providers object

kalxas:
The optional providers object in record is missing an affiliation/organization name property.
We do state that the required property `name` can also include the organization name, but it does not cover the affiliation case for the individual.


kalxas:
Description has been updated. For review.

Issue #171, opened 2022-06-21
Finish chapter 6 Overview


lvdbrink:
Chapter 6 of OGC API Records Core seems to still need some work: 
- some headings start with "tl;dr", this needs to be addressed
- there is a Note at the start about the need to add a discussion about migration from big repositories of metadata to distributed crawlable catalogues
- I suggest reordering the chapter. Start with what is now 6.4 General and 6.5 Building Blocks, integrate 6.1-6.3 into what is now 6.5, then the rest.
- the second and third paragraph of 6.8.2. API Behaviour Model could perhaps be deleted. As a reader I don't find they are necessary / don't contain need to know information.
- Tables 5 and 6 in chapter 6 are repeated in chapter 7 and further, in the conformance classes. In chapter 6 I think just a summary and a reference to the full table in the conformance class would be enough, e.g. "he core set of properties that may be used to describe a record collection or catalogue include id, title and description. See table xx for the full set."

pvretano:
08-AUG-2022 SWG Meeting: @pvretano will review and address.

pvretano:
Made the following changes:
(1)  Moved all the tl;dr clauses to an informative annex called "implemenation Summaries".
(2) I reordered the clauses as suggested BUT moved 6.1-6.3 to an annex as described in (1).
(3) Removed the last paragraph from the "API Behavioural Model" clause.  Paragraph 2 is highlighting the fact that CSW is based on a single metadata repository while OARec operates on any number of metadata repositories (i.e. collections).  That seems like useful information to me.
(4) Removed the tables from clause 6 a simply reference the appropriate tables of properties from the other clauses.

lvdbrink:
Thanks Peter, I'm happy with the changes!

Agreed about the Note; maybe we should start chapter 6 with a new paragraph titled "Role of OGC API Records" or something similar, where we explain what we envisage this role to be. I agree with the point that the Note makes.

I'll have a first go:

"Althought OGC API Records is built upon the legacy of CSW, its role in a modern SDI is a different one: catalogues will no longer be big repositories of metadata documents (e.g. ISO19115) like in traditional SDIs. Instead the architecture of a modern SDI is one of distributed, crawlable catalogies (e.g. static STAC). Resources are harvested into the catalogue to make them searchable. These resources can be metadata records close to resources, such as STAC items, or even collection objects."

... and then perhaps a few more sentences, I think we can probably get some content from the Abstract.

[...]

PR #199, opened 2023-01-27
Core specification review and editorial

kalxas:


kalxas:
During my review I took some notes for discussion. For example I see that one of my notes are already documented as issue #171 
Should I open issues for all points I recorded?

Issue #176, opened 2022-07-07
recordJSONFG schema missing?


lvdbrink:
In **Requirement 93 | /req/jsonfg/content** under C, there is reference to a schema `recordJSONFG.yaml`. 

But directly below, the included and displayed schema is not a JSON-FG schema, but a GeoJSON schema (the same as referenced in the GeoJSON requirements class directly before). 

There is no schema `recordJSONFG.yaml` that I could find in this github repository and am not sure of the intent here. 

If the intent is to include a JSON-FG schema here, then it needs to be added and the schema inclusion link should be corrected. 

If the intent is to use the same GeoJSON schema here as in the GeoJSON requirements class, then the text in requirement 93 C should be corrected.

pvretano:
@lvdbrink just a place holder for now.  If we decide to keep the JSON-FG conformance class then we can fill out the schema for a JSON-FG encoded record.

pvretano:
08-AUG-2022 SWG Meeting: We will leave the current content in the document until after the upcoming code sprint.  At the code srpint we will have the opportunity to explore JSON-FG and Records since both are topics of the code sprint.

cportele:
I had a look at the current text of the [JSON-FG requirements class](https://docs.ogc.org/DRAFTS/20-004.html#requirements-class-jsonfg-clause). Some questions:

* What should be the content of the JSON-FG "time" member? Is that up to the provider / the nature of the resource or should there be guidance based on the created/updated information about the record or its distributions?
* Should the "type" property be mapped to the JSON-FG "featureType" member?

In addition, see https://github.com/opengeospatial/ogc-feat-geo-json/issues/56.

pvretano:
I would say that the "time" value is up to the provider / the nature of the resource ... not the created/update dates of the record itself.
About "featureType" ... yes I think the type should be mapped to the "featureType" member.

cportele:
OGC Code Sprint 2022-09-15:

* Returning records in a CRS that is not WGS84 lon/lat is not an important use case for a Record API.
* Records specifies an application schema with standardized record properties, just like STAC. The same argument as stated in https://github.com/opengeospatial/ogc-feat-geo-json/issues/69#issuecomment-1248132912 applies for Records as well. We will therefore remove the JSON-FG requirements class from the current Records draft and include "time" and "type" in the "properties" member for now. If communities would start to use JSON-FG with record data we could revisit this in the future.

pvretano:
[...]

Issue #206, opened 2023-01-27
Add reference in class_jsonfg requirement


kalxas:
https://github.com/opengeospatial/ogcapi-records/blame/master/core/standard/requirements/requirements_class_jsonfg.adoc#L9

pvretano:
@kalxas since we decided a while back that jsonfg is not currently useful to Records, can we close this issue?

kalxas:
@pvretano shall we remove that line above then?

pvretano:
Closing. Resolved by #210 which removes the JSON-FG conformance class as per SWG decision made in issue #176.

kalxas:
Thank you

PR #210, opened 2023-02-20
Remove JSON-FG conformance class.

pvretano:
As decided by the SWG after the last code sprint, this PR removed the JSON-FG conformance class.


pvretano:
Closes issues #176 and #206.

Issue #195, opened 2022-12-12
Clarification on language(s)


m-mohr:
As far as I can see, `hreflang` is meant to follow RFC 5646 (Language-Tag). For the `language` property the format seems undefined. I'd propose to clarify that it uses the same format as `hreflang`.

Additionally, I'm wondering whether it would be helpful to define a list of available/supported languages, e.g. as a property `languages`, which is an array of languages.

Also, how should alternative representations in other languages be communicated in (static) catalogs? Maybe multiple `self` links with different `hreflang`s?

I'm asking because I'm writing this up for STAC and would like to align as much as possible.
See also https://github.com/stac-extensions/language and https://github.com/stac-api-extensions/language

m-mohr:
This was a very quick turnaround, thanks.

I'm confused on point 1: Why replace language with languages? I think both should exist:
1. language specifies the actual language you've received.
2. languages specified which languages are available (this is more or less a shortcut for checking all alternate links for the set of available hreflang's)

m-mohr:
Example from STAC:

```json
{
  "stac_version": "1.0.0",
  "stac_extensions": [
    "https://stac-extensions.github.io/language/v1.0.0/schema.json"
  ],
  "type": "Feature",
  "id": "item",
  "bbox": [...],
  "geometry": {
    "type": "Polygon",
    "coordinates": [...]
  },
  "properties": {
    "datetime": "2020-12-11T22:38:32Z",
    "example": "An example product",
    "languages": [
      "de",
      "en"
    ],
    "language": "en"
  },
  "links": [
    {
      "href": "https://raw.githubusercontent.com/stac-extensions/language/main/examples/item.json",
      "rel": "self",
      "hreflang": "en"
    },
    {
      "href": "https://raw.githubusercontent.com/stac-extensions/language/main/examples/de/item.json",
      "rel": "alternate",
      "hreflang": "de"
    },
    {
      "href": "catalog.json",
      "rel": "parent",
      "title": "Example STAC Catalog",
      "hreflang": "en"
    },
    {
      "href": "catalog.json",
      "rel": "root",
      "title": "Example STAC Catalog",
      "hreflang": "en"
    }
  ],
  "assets": {
    "data": {
      "href": "https://cloud.example.com/examples/file.tif"
    },
    "metadata": {
      "href": "https://cloud.example.com/examples/metadata.xml",
      "type": "application/xml",
      "hreflang": "en"
    },
    "metadata_de": {
[...]

PR #211, opened 2023-02-21
Update languge handling based on discussion with STAC.

pvretano:


core/openapi/parameters/language.yaml
```diff
@@ -0,0 +1,12 @@
+name: language
```
m-mohr, line 1: I thought you would go for a relatively loose "note" as in Features for the accept header, e.g. 

> Requirement X implies that the language of a server response is determined using content negotiation as specified by the HTTP RFC.
> 
> Note that any server that supports multiple languages will have to support a mechanism to mint encoding-specific URIs for resources in order to express links, for example, to alternate representations of the same resource. This document does not mandate any particular approach how this is supported by the server.
> 
> As clients simply need to dereference the URI of the link, the implementation details and the mechanism how the language is included in the URI of the link are not important. Developers interested in the approach of a particular implementation, for example, to manipulate ("hack") URIs in the browser address bar, can study the API definition.
>
> NOTE: A common approach is an additional query parameter (for example, "language" or "accept-language") that overrides the Accept-Language header of the HTTP request.

I'm fine either way though, we can also go for a specific `language` query parameter, but in this case the schema is not ideal I think. In #195 you asked that the parameter is aligned with the accept-language header. Although the schema mandates comma separated values, it's not clear from the schema that it can include the q thing etc. I assume it could be something simple like:

```yml
  type: array
  items:
    type: string
    pattern: "^([a-z]+)(-[a-zA-Z0-9]+)+(;q=(0|0.\d+|1|1.0+))?$"
```

If you don't go for the pattern, a description how this needs to look like (i.e. equivalent to the accept-language header) would be good.

core/openapi/schemas/language.yaml
```diff
@@ -0,0 +1,34 @@
...
+      The language tag as per RFC-5646.
+    example:
+      code: el
+    name:
+      type: string
+      description:
+        The untranslated name of of the language.
+      example:
+        name: Ελληνικά
+    alternate:
```
m-mohr, line 19: We recently got a long comment from the STAC community that "alternate" is pretty American and "alternative" would be the better alternative as it's better to understand. Thoughts? Should we switch to alternative?

See https://github.com/stac-extensions/alternate-assets/issues/9

m-mohr:
Hey @pvretano. Any thoughts from your side regarding the comments above?

core/openapi/schemas/language.yaml
m-mohr: I just realized that the rel-type is also alternate, so let's keep it consistent.

core/openapi/schemas/language.yaml
```diff
@@ -0,0 +1,34 @@
...
+  - code
+properties:
+  code:
+    type: string
+    description:
+      The language tag as per RFC-5646.
+    example:
+      code: el
+    name:
+      type: string
```
m-mohr, line 14: For name and alternate I've set `minLength: 1` in the STAC schema. Empty string don't seem overly useful.

m-mohr:
@pvretano I've created an OpenAPI document for the STAC extension that defines the headers and parameters more in detail, feel free to port over whatever you think is useful for you: https://github.com/stac-api-extensions/language/pull/2

m-mohr:
@pvretano Sorry, I don't want to nudge you by any means, but it would be very helpful for the STAC API work around languages if we could align our work until the 29th of March as I have to release until then. Thanks in advance.

pvretano:
@m-mohr my appologies ... keep getting distracted with other stuff  I will have reviewed your comments and made adjustments before the end of the week.  Please stand by ...

core/openapi/parameters/language.yaml
pvretano: @m-mohr I have updated the `langauge` parameter defined in `language.yaml` to match STAC.  Please review and let me know if it is better now.

I should point out that the definition of this parameter is not normative as the specification does not mandate a particular apprach for minting language specific URLs.  I realized, however, that I was not referencing that prameter schema (i.e. language.yaml) anywhere in the text of the specification so I now include it the the section titled "Langauge of the record" as an example of how a server can mint language specific URLs.

As for your loose note, the API for records is the Feature API so this is language that the features API should  include somehwere.  All records says, with regard to the API, is do what Features does.  Specifically, records says:

Implementations of the Records API SHOULD comply with recommendation [`/req/core/string-i18n`](http://docs.ogc.org/is/17-069r3/17-069r3.html#string_i18n) of the [OGC API - Features - Part 1: Core](http://docs.ogc.org/is/17-069r3/17-069r3.html) standard.

m-mohr: Looks good, thanks.

m-mohr:
> Another question came up when defining the OpenAPI: What is prioritized when both the language parameter and a accept-language header is provided? query paremeter or header?

@pvretano Any thoughts on this?

m-mohr (approved):

cportele:
> Another question came up when defining the OpenAPI: What is prioritized when both the language parameter and a accept-language header is provided? query parameter or header?

I would not specify this, this should be up to the server.

m-mohr:
I've added a note:

> **Note:** If both the header `Accept-Language` and the query parameter `language` are provided by a client,
> it is recommended to use the values provided in the query parameter.
> This is recommended to avoid issues with some clients (such as a web browser) that may send headers
>that are not under control of a user.
> Ultimately, it's up to the implementation to decide which source it prioritizes.

Thanks, Peter. I've published the STAC extension now.

kalxas (approved):

core/openapi/schemas/language.yaml
```diff
@@ -9,24 +9,24 @@ properties:
...
+    type: string
+    description:
+      The name of the language in another well-understood language,
+      usually English.
+    example: Greek
+  dir:
+    type: string
+    description:
+      The direction for text in this language.  Either left-to-right
+      or right-to-left.  If this member is not specified then the assumed
```
m-mohr, line 28: But why list it anyway then? You'll likely never encounter English in rtl. This was for informational purposes when you don't have a mapping in your client, so having it explicit helps to use the right direction without looking it up. I think as it is right now it could also be removed completely.

cnreediii:
One little nit: Chinese street signs, calligraphy, and some business cards and personal letters are use a vertical direction

m-mohr:
@cnreediii On the other hand, the dir was inspired by HTML, which I assume is stable and well-designed. And their dir tag just allows ltr and rtl.

m-mohr:
Right now this PR conflicts the the corresponding STAC extension due to the different default values of `dir`. The addition of btt and ttb is not so much of an issue, but the different default value will lead to a conflicting default behavior. The default in STAC is ltr. Not sure about a path forward yet, so just stating it here for now.

core/openapi/schemas/language.yaml
pvretano: @m-mohr I think the problem is that if someone specifies "Arabic" as the language but does not indicate a direction a default value of `ltr` would be assumes and that is not correct.  I have updated to the enum to add directions `ttb` ( to resolve [@cnreediii ](https://github.com/opengeospatial/ogcapi-records/pull/211#issuecomment-1495012337) comment), `btt` and `auto`.  The default is `auto` indicating that the client determines (somehow) the direction.  The value `btt` is added for completeness.  I don't know if there is an actual language that prints/displays `btt`; we can remove it later if it is useless.

pvretano: @m-mohr I'm going to merge this PR just to get it into the mainline.  If additional changes need to be made we can do that in another issue.

pvretano:
> Right now this PR conflicts the the corresponding STAC extension due to the different default values of `dir`. The addition of btt and ttb is not so much of an issue, but the different default value will lead to a conflicting default behavior. The default in STAC is ltr. Not sure about a path forward yet, so just stating it here for now.

@m-mohr isn't `auto` a safer default?

m-mohr:
@pvretano If the tooling can handle auto-detection, probably. But I'm not sure that's something we can expect.

pvretano:
@m-mohr we are not specifying how the client does `auto` detection.  `auto` just means that the client has to figure it out.

The client can -- for example -- simply assume `ltr` and go with that.  More sophisticated clients can do more ... use the `Content-Language` header and use internal knowledge to set the direction OR look up the language code and get the direction, etc.  The point is that `auto` simple means "You figure it out'"

pvretano:
@m-mohr I agree with your point but they equally well apply to just making `ltr` the default too ... no?  Is your comment implying that we should not set a default and instead simply make dir required and force data providers to set the direction?  I'd be OK with that too.

m-mohr:
My initial proposal was based on HTML dir, except that I added a default value while for HTML it is undefined.

cnreediii:
@m-mohr Thanks for the clarification. Should have read the entire thread :-)

pvretano:
@m-mohr @kalxas I made the default direction `ltr` and added a description that basically says that `ltr` is the default because that is the most common case.  I also said that care should be taken to set `dir` appropriately if your language directions is not `ltr`.  Pleave review.  If you are OK with the changes then I will merge this PR.

core/openapi/schemas/language.yaml
```diff
@@ -0,0 +1,33 @@
...
+    alternate:
+      type: string
+      description:
+        The name of the language in another well-understood language,
+        usually English.
+      example: Greek
+    dir:
+      type: string
+      description:
+        The direction for text in this language. The default ltr represents the most common situation.  However, care should be taked to set the value of dir appropriately if the language direction is not ltr.
```
m-mohr, line 27:
```suggestion
        The direction for text in this language. The default `ltr` (left-to-right) represents the most common situation. However, care should be taken to set the value of `dir` appropriately if the language direction is not `ltr`. Other values supported: `rtl` (right-to-left), `ttb` (top-to-bottom), and `btt` (bottom-to-top).
```

Just some formatting and maybe it's a good idea to explain what the individual values mean...

m-mohr (approved):
Thanks! Looks good to me. I just proposed some changes to formatting and descriptions to clarify a bit more.

Issue #215, opened 2023-03-03
Identifying different record types


ilkkarinne:
This is probably just me being fresh in the API Records spec, but how should one define and advertise different record types in API Records?

I'm working on an idea of defining extensions of the core queryables set of the properties for the records schema, such as properties from the (Geo)DCAT and implementations of the Observations, Measurements and Samples (OMS) concepts. That would probably mean using more than one record schema within a single API Records implementation (ObservationCollection record, ObservableProperty record etc.). I guess it would be clearest to allow only one record type (itemType?) per collection, but which properties should I use for describing the record schema used inside a particular record instance?

For the GeoJSON / JSON-FG format records the mandatory ```properties.type``` seems to be reserved for a natural language description of the _resource type_, not the record type. Should I use the top level ```featureType``` of the record GeoJSON feature like the following?
```json
{
   "type": "Feature",
   "id": "obscollrec-1",
   "featureType": "oms:ObservationCollectionRecord",
   "time": {
      "interval": [
         "2014-04-24T10:50:18Z",
         ".."
      ]
   },
...
```

If yes, how would I best declare the record type allowed for a particular collection in the collection metadata (the `itemType` is fixed to "record")?

ilkkarinne:
Thanks @pvretano, I think having an optional the `properties.recordType` to the core queryables would indeed be useful.

However, as knowledgeable client I would also like being able to determine the record type of records in a collection from the Collection schema. This way I could be prepared to parse the resultins of an `ìtems` query without testing the existence of `properties.recordType` value of each record separately. I do see the benefit of keeping the `itemType="record"` as a simple distinction from other collections (such as features), so could there be a `recordType` property also in the Collection schema?

pvretano:
@ilkkarinne I'll raise the recordType-in-the-collection-object in the SWG too...

pvretano:
06-MAR-2023:  SWG likes the idea of being able to indicate in a record which "extensions" are being used.  However, the SWG is not so keen on the `propertie.recordType` proposal.  Instead they would prefer we use an `extensions` mechanism similar to STAC.  So if, for example, a record is extended with GeoDCAT elements, the `extensions` section can list that the GeoDCAT extension was used.  (Have to run off to another meeting but will update this issue later today with additional details).

m-mohr:
Isn't this basically just a conformance class list but in the childs?
[...]

PR #216, opened 2023-03-31
Add a conformsTo member to the collection level.

pvretano:
This `conformsTo` member at the collection level allows extensions used in all catalogue records to be declared without having to repeat this information in every record.
Resolves issue #215.


pvretano:
03-APR-2023:  The standaridzation target in clause 2 and other placed needs to be updated to include "record" as well as API since the standardization target of the "conformsTo" property is not an API but the record.

PR #220, opened 2023-03-31
Add link to the queryables for local resources catalogue.

pvretano:


pvretano:
03-APR-2023:  May no actually need this PR is this generalization is done in features. @cportele has agreed to do that so we will leave this PR open until then and delete it once the changes have been made in features ... or reference that change in the Records specification

kalxas (approved):

Issue #227, opened 2023-04-12
Providers: Isn't the first schema just a link?


m-mohr:
When writing https://github.com/opengeospatial/ogcapi-records/issues/178#issuecomment-1505797658 I realized that the "first" schema (i.e. url and roles) is basically a weak variant of a link.

It is a URL with roles. This basically is the same as a href and a rel in a link, BUT it has the weakness that 
1. url != href
2. no `type` is specified

So the example usecase with the vcard (VCF) means a client can't properly detect what kind of file it actually loads. If the client would know upfront whether it is a VCF, it could actually do something useful with it.

Thus, I'm wondering whather the "first" schema should actually be removed and instead be a link with a "author" (or some kind of OGC URI such as `http://www.opengis.net/def/curie/ogc/1.0/rel/contacts`) rel type?

Context with regards to STAC: In STAC the `url` is meant to point to a webpage that contains browsable information, not necessarily a machine-readable format as implied by OAR. The intentions seem somewhat different so I don't think they necessarily need to align.

pvretano:
@m-mohr I've removed the first schema from party.yaml altogether since it is not really necessary.  The larger schema only has `name` as mandatory so you can make it as slim or fat as you like including only the bits of contact information desired.  This was implemented in PR #230.

kalxas:
I think this should be closed, the PRs are merged.

m-mohr:
Indeed.

Issue #103, opened 2021-02-26
contactPoint as uri?


pvgenuchten:
The current draft indicates to reference a contactPoint as uri. This prevents contactdetails to be repeated multiple times, with risk of typos. However I want to bring in some arguments to introduce explicit contact details in each record (or provide the option to choose, either uri or embedded).
- Not many organisations share their contact details as a vcard in a repository
- vcard is suggested, but if they use schema.org/Person, foaf or even unstructured html or ldap, clients will have a hard time parsing that information
- Organisation-name is an interesting queryable parameter

edpars0ns:
I would support schema.org/Person, indeed I think will should look at schema.org structures where possible - seems a key parameter from a user perspective is accessing the relative value of datasets.

pvretano:
@pvgenuchten how does this prevent contact details from being repeated multiple times with the risks of typos?  If anything wouldn't the opposite be true?  I have a single contact point object that lives somewhere.  I can then reference that contact object in records multiple times using a URL.  Seems to me that would prevent typos from creeping in.  No?
I think your second point if true for any format.  No? You suggest schema.org/Person but if I use VCARD, foaf, HTML, etc.  I will have a hard time parting that information.  That is why VCARD is a suggesting and not a requirement.  B.T.W. VCARD is what DCAT suggested so we went with that.

pvgenuchten:
@pvretano to clarify, I started with listing a benefit of the linkage to a remote resource. But I'm actually doubting if that benefit outweighs the negative aspects of such an approach.

If we suggest an ontology, it means not hardcoding it in to the standard, clients will have a hard time parsing it. So i actually propose to explicitly define the contact properties in the standard. Which ontology to use on top of that is more a conformance class issue, with json-ld-context you're probably able to expose the full record (including contact detail specifics) in the optimal ontology requested by the user.

pvretano:
@pvgenuchten LOL!  My bad ...  for some reason I though you were pointing that out as a bad thing.  Can you sketch up a quick JSON schema for the contact person property for the SWG to consider at the next meeting?  You can either create a PR or simply include your proposal in this issue. Perhaps you can include your initial proposal of it being either a URL or embedded content ... I like that idea.

pvgenuchten:
A proposal for a contact point (inspired by iso19139 & schema.org)

```json
{ "individualname": "",
  "organisationname": "",
  "url": "",
  "email": "",
  "phone": "",
  "postalAddress": {
[...]

PR #230, opened 2023-04-21
Change "providers" to "contacts" to align better with STAC, ISO19115, etc.

### LINKED ISSUE #228: Providers: Change to more generic contacts?
Moving the "providers vs contacts" discussion from https://github.com/opengeospatial/ogcapi-records/issues/178 to a separate issue:

> Question: what would the distinction between "providers" and "contacts" be in STAC? Maybe if I understand that I can propose changes to bring STAC and records closer together on this point.

So providers in STAC is really in the sense of "provided something" (reflected in the roles: licensor, processor, producer, host). And in this case really just the basics (url with additional details, name, role, description).

Contacts I'd see more as a general construct to provide contact details of related organizations or persons, which can include a lot more information. Could be a administrator, public relations manager, project officer, other contributors or maintainers, ... So this is far more advanced and less restricted than providers. And it will not be core, but an extension. 

Having that said I'd like to get to an alignment with between OAR and STAC that doesn't lead to problems in implementation due to differences. STAC clients would expect a `name`, could we require this in both schemas OAR allows? On the other hand, OAR would expect a url and that might not be present in STAC so OAR clients could also error. Not sure how to fix that. I assume a roles array with free-form roles is not ideal, but also shouldn't lead to too many issues downstream. 

So I'm wondering whether we should try to align this so that it doesn't lead to issues or split the definitions (which would mean OAR would need a different field name). I could see adopting the OAR definition for the contacts extension, but that would mean that OAR would need to switch to a "contacts" field name, which I feel is the better term anyway for OAR...

Related issue: #227

pvretano:
This change allows STAC to keep it's "providers" array as is and aligns Records with STAC, ISO19115 and other specifications that use the term "contacts" for this information.

Closes #228.


kalxas (approved):

kalxas (approved):

m-mohr:
Thanks, I'll take a look next week and try to create a STAC extension based on it and give feedback then.

pvretano:
21-APR-2023:  From the [party.yaml](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/party.yaml) schema, it originally was a `oneOf` that allowed either a reference to contact information or this big long object that provided explicit in-line information about the contact.  Problem is that for this "big long object" the only required field was `name` and it also included the ability to link out to information about the contact.  So why have a seperate object for references?  If someone wants to "reference" contact info, they simply provide the `name` and the link information (in `contacts.contactInfo.url`) and it provides exactly the same capability.

Closes #227

pvretano:
Closes #103

tomkralidis (approved):

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
-              responsible party.
-      roles:
-        $ref: roles.yaml
+              The key name indicates the type of email address (e.g. home,
+              work, etc.). The value of the email address itself.
+            format: email
+      address:
+        type: object
+        description: Physical location at which contact can be made.
+        patternProperties:
```
m-mohr, line 53: This line and the one below is equivalent to `additionalProperties`, no?

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
-              Time period (including time zone) when the resposible party
-              can be contacted.
-          contactInstructions:
+              The key name indicates the type of phone number (e.g. home,
+              work, fax, etc.). The value is the phone number itself.
+            pattern: "^+[1-9]{1}[0-9]{3,14}$"
+      email:
+        type: object
+        description: Email address at which contact can be made.
+        patternProperties:
```
m-mohr, line 43: Same questions as for address.

core/openapi/schemas/party.yaml
tomkralidis: I think the objects are to provide keys, so:

```json
                   "phone": {
                        "office": "+01-819-997-2800"
                    },
                    "email": {
                        "main": "you@example.org",
                        "personal": "foo@example.org"
                    },
```

cnreediii:
OK, I have to ask . . . What is the relation (or not) to vCard (https://datatracker.ietf.org/doc/rfc6350/). And if no relation or possibility of using, why not? A vCard has a .VCF extension and is supported by Microsoft Outlook (98-2019), Mozilla Thunderbird as well as most mobile clients. And governments such as the UK recommend using vCard. Which makes it a surprisingly well supported format for storing contact info - even though some folks consider it obsolete.

pvretano:
@cnreediii the information about a contact can be encoded in-line in the record OR referenced.  If it is referenced, the specification recommends that it reference a vCard.  I don't think anyone considered actually encoding the vCard inline in the record.  A change like that would need to be discussed in the SWG but it would certainly make the specification shorter since we could just reference the RFC.

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
+  positionName:
+    type: string
+    description: Role or position of the responsible person.
+  organization:
+    type: string
+    description: Organization/affiliation of the individual/responsible person. In case of an organization, the name property should be used and this property is not to be used.
+  logo:
+    $ref: common/link.yaml
+    description: Graphic identifying a party
+  contactInfo:
```
m-mohr, line 25: I think it would make sense to simplify this a little and move the fields defined in the contactInfo one level up. 
I'm aware that this is coming from ISO 19115, but it feels like a XML thing. It's a 0..1 relation, so there's no actual functionality behind it, e.g. to allow multiple of these objects. What do you think?

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
+              administrativeArea:
+                type: string
+                description: State or province of the location.
+              postalCode:
+                type: string
+                description: ZIP or other postal code.
+              country:
+                type: string
+                description: Country of the physical address.
+      url:
```
m-mohr, line 75: I think the name "url" is a bit misleading. This is actually an object, not just a url. Maybe rename to link or so? ISO calls it onlineResource (but I don't like that).

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
+        $ref: common/link.yaml
+        description: On-line information about the responsible party.
+      hoursOfService:
```
m-mohr, line 78: I don't think a free-text is very useful here. Sure, you can add everything, but it's not machine readable at all. I'd prefer to either remove this or make it more structured. Thoughts?

core/openapi/schemas/party.yaml
m-mohr: Okay, I don't think it's a good idea thoigh. For example, you can't translate this.

If you use an array with a role (or so) you can change the values depending on the language, but if you change the key, you may break implementations as they may directly access entries through the key.

For example, if you have a multi-language catalog, in english you have:
` "phone": { "office": "+49123456789" },`
In German it gets ugly:
` "phone": { "Büro": "+49123456789" },`

So as long as they keys are not pre-defined and users can put whatever they like, I think this should not be in the keys.

core/openapi/schemas/party.yaml
m-mohr: As with logo, which relation type should be used? I assume there are different use cases (e.g. homepage of company vs. vCard), but a bit of guidance would be appreciated. It's required so people will stumble across it (as I do right now when creating an example file).

cnreediii:
@pvretano Thanks. Should have been more clear. I was also thinking of vCard by reference and not inline. Same as license discussion. Missed that the draft already suggested vCard. Sigh

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
+        type: object
+        description: Physical location at which contact can be made.
+        patternProperties:
+          "^.*":
+            type: object
+            description: |-
+              The key name indicates the type of address (e.g. office,
+              home, etc.). The value is the address itself.
+            properties:
+              deliveryPoint:
```
m-mohr, line 60: Deliverypoint in ISO 19115 is 0..n, here it's 0..1 - intentional?
How do I specify multiple lines? Can I use line-breaks?

m-mohr (changes requested):
Several comments above/below that we should discuss. The pattern issue should be solved at least.

fyi, here's an actual example file based on this schema:
https://github.com/stac-extensions/contacts/blob/main/examples/collection.json

core/openapi/schemas/party.yaml
pvretano: Agreed.

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
+    description: |-
+      The name of the organization or the individual.
+  positionName:
+    type: string
+    description: Role or position of the responsible person.
+  organization:
+    type: string
+    description: Organization/affiliation of the individual/responsible person. In case of an organization, the name property should be used and this property is not to be used.
+  logo:
+    $ref: common/link.yaml
```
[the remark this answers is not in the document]
pvretano: @m-mohr added a requirement that says if the logo member is included then its link relation should be `icon` and its media type should be an image media type.

core/openapi/schemas/party.yaml
pvretano: @m-mohr made deliveryPoint an array to add multiple address lines.

core/openapi/schemas/party.yaml
pvretano: @m-mohr change the name of `url` to links and made it an array.  Also added requirement that all links in the `links` list shall use the rel `about` and include a media type indicating the content type of the link (e.g. text/html for a company's web page versus text/vcard for virtual contact file).

core/openapi/schemas/party.yaml
```diff
@@ -1,103 +1,89 @@
...
+        patternProperties:
+          "^.*":
             type: string
             description: |-
-              Time period (including time zone) when the resposible party
-              can be contacted.
-          contactInstructions:
+              The key name indicates the type of phone number (e.g. home,
+              work, fax, etc.). The value is the phone number itself.
+            pattern: "^+[1-9]{1}[0-9]{3,14}$"
```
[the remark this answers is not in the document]
pvretano: Fixed.

core/openapi/schemas/party.yaml
pvretano: @m-mohr threw together a schema for HOS.  Not sure I did it correctly or usefully but the idea is the be able to say something like:
````
hoursOfService: {
   "1": [ ["9:00","17:00"]], 
   "3": [["9:00","17:00"]],
   "4": [["10:00","12:00"],["13:00","15:30"]],
   "closed": ["-12-25","-07-20"]
}
````
Please review and mark as resolved if you are OK with the changes.

tomkralidis: fwiw in pygeoapi we've loosely aligned with https://schema.org/openingHours

pvretano: @tomkralidis, yeah, saw that.  Was not crazy about it.  Didn't think it was granular enough but if people like that instead, we can go with that ... or remove HOS altogether.  Do people really use this?

tomkralidis (commented):
Do we need both providers and contacts?  I would imagine contacts would be enough?

core/openapi/schemas/party.yaml
tomkralidis: We use this in MSC.

pvretano:
@tomkralidis added it because @m-mohr suggested we might consider adding it for STAC alignment.  I believe that STAC is planning to include both providers and contacts (as an extension).  It is my understanding that providers are related to the creation, processing or storing of the the resource(s) that the record is describing.  Contacts are other related parties but not involved in the creation, processing or storage of the resource directly.  The distinction seems a little subtle to me but we can easily remove providers if people think is it superfluous.

m-mohr:
Yeah, you can also offer it to your users, but I think on Friday you said you would add it as extension and I think that's reasonable. providers doesn't need to be in the core. 

Thanks for all the updates, @pvretano. I'll look into the other reviews and comments in the next days, but now I first need to get rid of COVID... :-(

kalxas (approved):

pvretano:
27-APR-2023:  party.yaml will be renamed to contacts.yaml after merge.

core/openapi/schemas/party.yaml
tomkralidis: @pvretano @m-mohr looking at this again, isn't`hoursOfService` as  a `string` enough here?  I'm just concerned that providers will have issues producing a valid definition.  Yes, a text string is relativiely free form and perhaps not as machine readable, but it is human readable and fits the use case of review/assessment of metadata when looking to get in touch with a contact.

m-mohr: Yeah, this was merged rather quickly. I think the current definition is not ideal ("1sad" is a valid key, isn't it?), but a string is also not ideal. I feel like hoursOfService is so niche, that I'd actually remove it as it can also be encoded in the free-text field `contactInstructions` as it is describes as "[...] instructions on [...] when to contact the responsible party."

core/openapi/schemas/party.yaml
m-mohr: Moved to #243

core/openapi/schemas/party.yaml
m-mohr: Created #245 for it. Maybe you want to add your comment there, as well? @tomkralidis

PR #240, opened 2023-04-27
add annex for contact role codes

tomkralidis:
As discussed at the OGC/OSGeo/ASF joint sprint.


kalxas (approved):

kalxas:
Yes, I think pointing to those links as W3C is doing in https://www.w3.org/TR/vocab-dcat-2/#Class:Role is the best option here.

kalxas:
We should probably do something similar with type:
https://www.w3.org/TR/vocab-dcat-2/#Property:resource_type

Issue #235, opened 2023-04-25
Confusing schemas for `themes`


m-mohr:
I just searched for information about `themes` and found various schemas and examples, which was a bit confusing:

Maybe check the following files whether they are up-to-date:
- themes is an array of URIs? Looks wrong: https://github.com/opengeospatial/ogcapi-records/blob/master/core/standard/clause_12_local_resources_catalogue.adoc#example
- type: datetime? What's that? https://github.com/opengeospatial/ogcapi-records/blob/master/core/examples/json/queryables.json

These look consistent:
- https://github.com/opengeospatial/ogcapi-records/blob/master/core/examples/json/record.json
- https://github.com/opengeospatial/ogcapi-records/blob/master/core/examples/json/keywords-themes.json
- https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/catalogue.yaml
- https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/recordGeoJSON.yaml
- https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/ogcapi-records-1.yaml

pvretano:
@m-mohr yeah, the examples are not quite up to date relative to the schemas.  I have it on my todo list to update and augment them.

m-mohr:
@pvretano Thanks. The "good" schemas itself also have an issue, I think. Neither `concepts` nor `scheme` is required. I assume both should be required?

In the STAC extension I made the schema a bit more strict in general:
```json
{
          "type": "array",
          "minItems": 1,
          "items": {
            "type": "object",
            "required": [
              "concepts",
              "scheme"
            ],
            "properties": {
              "concepts": {
                "type": "array",
                "items": {
                  "type": "object",
                  "required": [
                    "id"
                  ],
                  "properties": {
                    "id": {
                      "type": "string",
                      "minLength": 1
                    },
                    "title": {
                      "type": "string",
                      "minLength": 1
                    },
                    "description": {
                      "type": "string",
                      "minLength": 1
                    },
                    "url": {
                      "type": "string",
                      "format": "uri"
                    }
                  }
                }
              },
              "scheme": {
                "type": "string",
                "minLength": 1
              }
            }
          }
        }
```

pvretano:
[...]

PR #246, opened 2023-04-30
fix up / make consistent themes definitions

tomkralidis:
Follow-on of #237, as per https://github.com/opengeospatial/ogcapi-records/issues/235#issuecomment-1528834725

- cleanup of queryables examples based on core queryables
- minor editorial updates/fixes

I also refactored the themes definition into its own schema fragment/file for reuse.  In theory we could use this in https://github.com/opengeospatial/ogcapi-records/blob/master/core/standard/clause_12_local_resources_catalogue.adoc#example, but the example would then show a schema reference rather than the actual inline definition (which is valuable from a spec readability perspective, but another place to update if things change i the themes definition).

Thoughts @pvretano @m-mohr?


core/examples/json/queryables.json
```diff
@@ -5,68 +5,68 @@
...
-            "queryable": "publisher",
-            "type": "datetime"
+            "queryable": "description",
+            "type": "string"
         },
         {
-            "queryable": "themes",
-            "type": "datetime"
+            "queryable": "keywords",
+            "type": "string"
```
m-mohr, line 33: How's the type determined? Isn't that actually an array of strings? Maybe I'm missing something though...

m-mohr:
I guess a reference would be okay and simpilfy the example. Or is there any good reason why in practice someone wouldn't use a ref and reuse existing schemas?

tomkralidis:
@m-mohr I've updated the example to match the queryables schema as well as the various types.

As well:
> I guess a reference would be okay and simpilfy the example. Or is there any good reason why in practice someone wouldn't use a ref and reuse existing schemas?

For an implementation, yes we would ref and reuse existing schemas. The question is whether to use a `$ref` in the example in the specification or be explicit to demonstrate the concept?

pvretano:
01-MAY-2023: Closes #235

m-mohr:
My review is above and did not receive comments yet.

core/examples/json/queryables.json
tomkralidis: I've updated the type based on the type of the actual queryable.  For example, if `keywords` is an array, the queryable is of type `array`.  Users searching would pass a string literal (`keywords=birds`).  @pvretano is this the right interpretation of defining queryables?

m-mohr: That makes sense, I've just seen other implementations (in STAC?), where the queryables are 1:1 mappings to the actual "property" (here array). This left me confused. If this PR is the correct way forward, great.

core/openapi/schemas/catalogue.yaml
pvretano, line 1: @tomkralidis the schema for the `themes` member should be:
````
themes:
  type: array
  description: A knowledge organization system used to classify the catalogue.
  items:
    $ref: theme.yaml
````
The PR I recently merged created a `theme.yaml` file that is used in few places.

core/openapi/schemas/themes.yaml
pvretano, line 1: @tomkralidis remove this file and verify that the [`theme.yaml`](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/theme.yaml) file that is now in the master branch contains the same information for a single theme.

core/openapi/schemas/catalogue.yaml
tomkralidis: @pvretano fixed/rebased.

core/openapi/schemas/themes.yaml
tomkralidis: FYI I'm now following what is in master, which means `themes.yaml`.

tomkralidis: Wait, I see the following now in master:

https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/recordGeoJSON.yaml#L110

...but there is no `themes.yaml`, only `theme.yaml`.

Summary: I've updated this PR to point to `theme.yaml`.  Should be all aligned now.

pvretano (approved):
@tomkralidis looks good.  Do we still need the `themes.yaml` file?

tomkralidis:
@pvretano good catch, removed.  Thanks

P.S. if/once this PR is merged, it should be "Squash and Merge"d (lots of superfluous commits).

m-mohr (approved):
Looks like a good improvement

PR #248, opened 2023-05-02
Change roles from dictionary keys to a key named 'role'.

### LINKED ISSUE #243: Use roles instead of key for "type" of phone number / email / ...
_Originally posted by @m-mohr in https://github.com/opengeospatial/ogcapi-records/pull/230#discussion_r1176582181_

I don't think it's a good idea to encode the "role" of a phone number / email / ... type in the key. For example, you can't translate this.

If you use an array with a role (or so) you can change the values depending on the language, but if you change the key, you may break implementations as they may directly access entries through the key.

For example, if you have a multi-language catalog, in english you have:
` "phone": { "office": "+49123456789" },`
In German it gets ugly:
` "phone": { "Büro": "+49123456789" },`

So as long as they keys are not pre-defined and users can put whatever they like, I think this should not be in the keys.

Better could be something like:
` "phone": [{ "value": "+49123456789", "roles": ["Büro"] }],`
 It would also allow assigning multiple roles per "value" (e.g. a phone number that is for home and work).

pvretano:
Change the role dictionary key to a key named `role`.
Thus something like this:
````
"phones": {
   "home": "+1-416-555-5555",
   "office": "+1-416-555-5554"
}
````
changes to something like this:
````
"phones": [
  { "value": "+1-416-555-5555", "role": "home" },
  { "value": "+1-416-555-5554", "role": "office" }
]
````

Closes #243

### CODE REVIEW

**Comment by pvretano** [comment]
@m-mohr for some reason you don't show up in the reviews list so I am adding this comment to ping you to review.

**Review by kalxas — APPROVED** [review]

**Comment by m-mohr** [comment]
I think I'd prefer "roles" as array, aligned with the contacts / providers. It tackles the use case where users have the same work and provivate addresses (as I have as a self-employed for example). I think that would be more versatile. But if others prefer the string strongly, I'll not block it.

**Comment by tomkralidis** [comment]
Since we use a `roles` array elsewhere (contact role), +1 to have a `roles` array here as well for consistency.

**Comment by pvretano** [comment]
@m-mohr @tomkralidis just to make sure, you want it to look like this ... yes?
````
"phones": [
  { "value": "+1-416-555-5555", "roles": [{"name": "home"}] },
  { "value": "+1-416-555-5554", "roles": [{"name": "office"}] }
]
````
Works for me, just verify this is what you have in mind and I'll update the PR.

**Comment by m-mohr** [comment]
I thought like this:

> ```
> "phones": [
>   { "value": "+1-416-555-5555", "roles": ["home"] },
>   { "value": "+1-416-555-5554", "roles": ["office"] }
> ]
> ```

> Works for me, just verify this is what you have in mind and I'll update the PR.

**Comment by pvretano** [comment]
@m-mohr ah OK.  I thought @tomkralidis was implying the reuse of the `role.yaml` schema in the master branch.  No worries.  I can make it an array of string ... works for me.

**Comment by tomkralidis** [comment]
Roles [as defined in contact objects proper](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/role.yaml) are consistent with the following:

```json
"phones": [
  { "value": "+1-416-555-5555", "roles": [{"name": "home"}] },
  { "value": "+1-416-555-5554", "roles": [{"name": "office"}] }
]
```

Having said this, no objections to moving `roles` to an array, however we should be consistent with contact roles as well (changing `role.yaml` so that it can be used in `contact.roles` and `contact.phones`, etc.  I would also tighten roles to have `minItems: 1` so that if someone specifies roles, that it should not be empty.

m-mohr:
Oh, I thought the roles in OAR were the same as the roles in STAC (assets + providers). Do you have any reason why you are not just using an array of strings in general, also for contact roles? I'm also just aiming for consistency here, but were caught in my STAC world ;-)

tomkralidis:
I think the idea of having a role (for contacts) as an object is being able to qualify (with a codelist) to provide further meaning to the value.  Different metadata standards have different vocabularies for roles, and in OARec we do not constrain which (contact) roles are put forth; we do provide an annex with known vocabularies (including ISO, DCAT, STAC, for example) in case users do not have one in place.

If it's important to keep consistent the structure with `contact.roles`, then we should keep with roles as an object.

Having said this, what about:

```json
{
  "phones": [
    {
      "value": "+1-416-555-5555",
      "type": "home"
    },
    {
      "value": "+1-905-555-5555",
      "type": "office"
    }
  ]
}
```

I know this introduces the possibility of having to duplicate a phone number, email, or address if we want to express multiple roles, but I'm not sure how often we will see the same contact type with multiple phone numbers/emails/addresses.  As well, given the OGC API - Records "Record" building is meant to provide a baseline, perhaps the above is enough?

m-mohr:
Generally, I feel like roles with codelists seems overengineered to me and nothing for core. If at all, it feels more like an extension. 
I guess type is fine, but then I still need to resolve the issue that roles != roles in STAC (contacts vs. providers). I don't like that roles in contacts is much more complex than roles in providers. I must admit I've not realized that the roles in contacts were so complex.

pvretano:
@m-mohr no problem with renaming `roles` to 'types` and making it an array of strings.  I'll make that change here. 

As for `roles` in general -- not sure how the others feel -- but I would prefer alignment with STAC in this area.  That would mean making all `roles` members arrays of strings rather than objects.

If organizations or communities of interests want the role values to be from some authoritative list then they can create a standard or best practice or guideline indiciating this requirement.  Even more.  Say there is a community of interest like WMO.  They develop and publish guidelines about how to use OGC API Records.  In those guidelines they can indicate that the value of the `roles` member(s) must come from some authoritative code list.  Hopefully in those guidelines they also define a `conformsTo` URL so that clients that recognize that URL will know that the record conforms to their guidelines and will thus also know that the value(s) of the `roles` member are taken from the specific, authoritative code list in the WMS guidelines.  You get the idea.

If the others (@tomkralidis, @kalxas) are OK with this approach I can add those changes to this PR.

m-mohr:
What you'd do in STAC for the authoritive roles, is probably eitehr what @pvretano proposes or to just re-use themes in contacts or providers. We have this notion of "common metadata" where you can use fields in many different places.  

If we align roles with STAC, we don't need to rename to types though...

m-mohr:
Just to clarify: "becomes an array" means "becomes an array of strings", right?
Sounds reasonable to me.

tomkralidis:
Correct, yes.

pvretano:
Question ... what is the difference between the `postitionName` and a `role` in the contacts strcuture?

tomkralidis:
from 19115:

- positionName: role or position of the responsible person
- role: function performed by the responsible party

pvretano:
@m-mohr @tomkralidis please review latest change an let me know if they are satisfactory.

m-mohr (approved):

kalxas (approved):

tomkralidis (approved):

PR #253, opened 2023-05-05
Add Role of OGC API Records in Overview chapter (#171)

### LINKED ISSUE #171: Finish chapter 6 Overview

kalxas:
Closes #171


pvretano (approved):

core/standard/clause_6_overview.adoc
```diff
@@ -1,7 +1,14 @@
 [[overview]]
 == Overview
 
-NOTE: Need to have a disucssion in here about the migration of catalogues from big repositories of metadata documents (e.g. ISO19115) to distributed, crawlable catalogies (e.g. static STAC) and harvesting of resource into the catalogue to make them searchable (e.g. harvesting records close to resources (i.e. STAC item) or even harvesting the resource itself (e.g. the collection object)).
+[[records-role]]
+=== Role of OGC API Records
```
tomkralidis, line 5: `s/OGC API Records/OGC API - Records/g`

core/standard/clause_6_overview.adoc
```diff
@@ -1,7 +1,14 @@
...
+
+Historically, catalogues have been used as big repositories of formal metadata like ISO-19119, ISO-19115, FGDC. etc. However, more recently catalogues are migrating to a more distributed model where the metadata is intimately associated with the resource that it describes. A modern example is STAC which relies on well defined building blocks that can be linked in different ways.
+
+Althought OGC API - Records is built upon the legacy of CSW, its role in a modern SDI can be both: a big repository of metadata documents or a distributed (even crawlable) catalogue.
```
tomkralidis, line 9: Although

core/standard/clause_6_overview.adoc
```diff
@@ -1,7 +1,14 @@
...
+
+In the modern scenario, the catalogue records can be distributed and linked together so that they can be crawled OR they can be harvested into a service endpoint to make them searchable. Even more than that, a searchable catalogue can harvest not only a metadata document describing the resource but in some cases the resource itself (e.g. harvesting the collections object from OGC API). The implications of this are subtle but important. In the traditional cataloging approach (e.g. based on CSW) a client could find the record and the record might include binding information to access the resource but that is pretty much it. Using the new approach, a client interacts more intimately with the catalogue and the catalogue becomes integral to the access to the resource. Think of the entire cloud native ecosystem: range requests to COGs don't really work unless there is a STAC item telling the client in what ranges the bands exists, thus the catalogue becomes an integral component in the process of accessing a resource.
```
tomkralidis, line 11: `s/Even more than that/In addition/`

tomkralidis: As well:
Replace:
Think of the entire cloud native ecosystem: range requests to COGs don't really work unless there is a STAC item telling the client in what ranges the bands exists, thus the catalogue becomes an integral component in the process of accessing a resource.

with

An example is the cloud native ecosystem: HTTP range requests to Cloud Optimized GeoTIFF (COG) data cannot be used propertly unless there is a STAC item telling the client in what ranges the bands exists, thus the catalogue becomes an integral component in the process of accessing a resource.

tomkralidis (changes requested):

core/standard/clause_6_overview.adoc
tomkralidis: Update to:

Historically, catalogues have been used as large repositories of formal metadata like ISO 19119, ISO 19115, FGDC, etc. However, more recently catalogues are migrating to a more distributed model where the metadata is intimately associated with the resource that it describes. A modern example is STAC which relies on well defined building blocks that can be linked in different ways.

core/standard/clause_6_overview.adoc
kalxas: Fixed, thanks

core/standard/clause_6_overview.adoc
kalxas: Fixed, thanks

core/standard/clause_6_overview.adoc
kalxas: Fixed, thanks

PR #260, opened 2023-05-15
Add Folder conformance class.

### LINKED ISSUE #222: Add the concept of the STAC Catalog to the crawlable catalogue clause
STAC has STAC Items, STAC Catalog and STAC Collection.  OGC API Records only describes Collections (i.e. catalogues) and Items (i.e. record).  However, if you are trying to describe a static or crawlable catalogue that contains a hierarchy that may not be possible without the idea of a "STAC Catalog".  This issue is created to explore this requirement a bit further.

pvretano:
Closes #222


kalxas (approved):

tomkralidis (approved):

core/openapi/schemas/folder.yaml
```diff
@@ -0,0 +1,61 @@
...
+    description:
+      An identifier of this folder.
+    type: string
+  type:
+    description:
+      Fixed value of "folder".
+    type:
+      string
+    enum:
+      - Folder
```
m-mohr, line 18: If this would be "Catalog" the same entity could serve STAC and Records. With this proposal you need two entities side-by-side, which I think is not ideal. (Or would they never touch anyway?)

pvretano: No, the intention is that they would touch but the problem is that "catalogue" in OAPIR means a "collection or records" so we had a terminology problem and decided to use the term "folder" since it seemed like the function of a "STAC catalogue" was to act like a folder or directory for organizing records to be crawled (hopefully that interpretation is correct) .

A folder is a collection or references to any type of OAPIR resource including records, catalogues, other folders.  A collection (or record collection  or catalogue in OAPIR) strictly references records.

Folder has the following members:
- id, type(=Folder), title, description, keywords, created, updated, language, languages, extent, links.

Record collection had the following members:
- id, type(=Folder), title, description, keywords, created, updated, language, languages, extent, links.
- itemType, crs, conformsTo, themes, contacts, licence, right, properties.

Originally "folder" had 4 members: id, type(=Folder), title and links but then we got into a bit of a "it would nice if" game and started adding more stuff! LOL!

m-mohr: Thanks, shifting the type discussion to https://github.com/opengeospatial/ogcapi-records/issues/264

m-mohr:
What's the actual difference between Catalog/Folder and Collection in OAR? The only required field that the Collection adds is really the title (for which many implementations just use the id). All other fields seem to be aligned (or missing). As most fields in Collections are optional anyway you could simply leave them out if you want to create a "folder" using a Collection, I don't clearly see what OAR adds a Folder for. You can pretty much can not distinguish them... In STAC a Collection requires more fields (license, extent), so has somewhat a reason to exist, but I don't quite get it in OAR. Is the only difference the type (and title) so that you can distinguish them in the API so that they are not listed in /collections?

pvretano:
29-MAY-2023:  The main sticking point right now is the value of the `type` field for the folder object.  Right now that value is fixed to "Folder" but this then introduces a friction point with STAC where the value of type for the analogous object is "Catalog".  We discussed this issue in the SWG and could not come to a consensus.  The options discussed were:
(1) Leave the value as "Folder".  This, unfortunately, forces people deploying crawlable catalogues that want to conform to this specification but also leverage STAC tooling to maintain two parallel structures (i.e. an OAPIR folder and a STAC catalog).
(2) Change the value of type to "Catalog".  This would allow someone deploying an OAPIR-compliant crawlable catalogue to leverage STAC tooling but would be a bit confusing since we use the term "folder" everywhere in this conformance class but the type is "Catalog".  I should point out that for this option we would include a note explaining why the conformance class is called "Folder" but the type is "Catalog".
(3) Allow either or value.  A possibility but introduces more option and thus more opportunities for incompatability.  

It would be good if @m-mohr could join the next SWG teleconference to add his thoughts to this dicsussion.

There was also mention of how difficult it would be for STAC tooling to be updated to accept either "Catalog" or "Folder".

m-mohr:
@pvretano On June, 12 I'm attending the OGC API - Tiles sprint remotely and on June, 26 I'm in Kosovo for FOSS4G, so it will be hard for me to join the Record meetings.

My long-term vision would be to make STAC a "profile" or "extension" of Records, but for that we need to at least erase all conflicts. The only real conflict I found so far (I'm currently [writing a comparison](https://github.com/stac-utils/stac-crosswalks/blob/master/ogcapi-records/README.md)) is the Folder type. I understand that the Catalog terminology is not ideal for Records, but on the other hand it's just a constant value in a JSON document that end-users would ideally never see (due to client software). Due to the broad adoption of STAC, I'd argue OAR1 has a good enough reason to align with STAC here.

cnreediii:
@m-mohr COncur: Write Records the way the SWG members think it should be and then craft one or more profiles/extensions.

pvretano:
Closes #260 
Closes #264

pvretano:
07-JUN-2023:  It was pointed out at the Records session at the TC that the term "Folder" might imply that duplicate entries are not allowed ... that is things cannot live in two folders simultaneously and thus we should think of a better name.  I am not sure I agree with this but I thought I should post the comment for consideration by the group. 
I found these on the Web: 

- https://simple.wikipedia.org/wiki/Folder_(computing)
- https://en.wikipedia.org/wiki/Directory_(computing)#Folder_metaphor

So, my feeling is that the term "Folder" is just fine.  However, if others have a different opinion please chime in.

m-mohr:
I heard that comment from @matthewhanson in the session. It was more an explanation why it's not called folder in STAC, but instead it's called Catalog. Folder sounds file system specific and people may thing that it's always a folder in the file system, which often it is not. A more generic term is Group, but that's probably too generic. Catalog was a good compromise. But I guess this issue is more in the terminology that is used in the specification and in spoken language. The actual conflict between STAC and Records is not there, it's just in the `type` field value. I don't think it necessarily need to match and even if the type is Catalog as in STAC, you can still name the whole construct differently.

cnreediii:
@m-mohr Well, the Wikipedia definition states, "a folder (also called directory, or catalog) ". Just sayin'

pvretano:
05-JUL-2023: Originally the Records specification had two objects ... the **collection** and the **record**.  If you wanted to encode a hierarchy ... as per the clause [Collection of catalogue resources](https://docs.ogc.org/DRAFTS/20-004.html#_collections_of_catalogue_resources) ... you simply used intermediate collection objects to link the resources up accordingly.  This seems to us a cleaner and more "compatible with STAC" approach then defining a new Folder conformance class.
So, @pvretano will (a) discard this folder conformance class and (b) clean up the language is the [Collection of catalogue resources](https://docs.ogc.org/DRAFTS/20-004.html#_collections_of_catalogue_resources) clause to be more clear.

pvretano:
No longer relevant since we decided to overload the record collection object to perform this function.

PR #262, opened 2023-05-20
schema updates

tomkralidis:
Addresses #263 

cc @pvretano @kalxas @m-mohr

### DIFF
--- a/core/openapi/schemas/contact.yaml
+++ b/core/openapi/schemas/contact.yaml
@@ -44,12 +44,14 @@ properties:
     description: Telephone numbers at which contact can be made.
     items:
       type: object
-      required: value
+      required:
+        - value
       properties:
         value:
           type: string
           description: The value is the phone number itself.
-          pattern: "^+[1-9]{1}[0-9]{3,14}$"
+          pattern: "^\\+[1-9]{1}[0-9]{3,14}$"
+          example: "+14165550142"
         roles:
           description:
             The type of phone number (e.g. home, work, fax, etc.).
@@ -59,7 +61,8 @@ properties:
     description: Email addresses at which contact can be made.
     items:
       type: object
-      required: value
+      required:
+        - value
       properties:
         value:
           type: string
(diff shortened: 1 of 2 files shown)


core/openapi/schemas/recordGeoJSON.yaml
```diff
@@ -23,7 +23,7 @@ properties:
     $ref: time.yaml
   geometry:
     oneOf:
-      - type: null
+      - type: "null"
```
cportele, line 26: This is an OpenAPI 3.0 schema according to the path, so you cannot use `null` as a type. The property needs to be annotated as `nullable: true` instead.

core/openapi/schemas/contact.yaml
```diff
@@ -44,12 +44,13 @@ properties:
...
       type: object
-      required: value
+      required:
+        - value
       properties:
         value:
           type: string
           description: The value is the phone number itself.
-          pattern: "^+[1-9]{1}[0-9]{3,14}$"
+          pattern: "^\\+[1-9]{1}[0-9]{3,14}$"
```
m-mohr, line 53: Weird. The change is correct, but we had another PR recently that fixed this and I'm wondering why the change got lost...

core/openapi/schemas/recordGeoJSON.yaml
tomkralidis: thanks @cportele.  Fixed.

m-mohr:
Setting all the GeoJSON schemas to nullable looks a bit weird and is probably not what we want. 

The oneOf could be fixed as follows, I think

```
    oneOf:
      - nullable: true
         enum:
           - null
      - $ref: geometryGeoJSON.yaml
```

tomkralidis:
Additions:
- added example for contact phone number
- simplified contact hours of service to be a plain string

cportele:
@m-mohr 

> Setting all the GeoJSON schemas to nullable looks a bit weird and is probably not what we want.

The change by @tomkralidis is correct. Yours is not a valid OpenAPI schema.

> But maybe Features has a better solution? I assume they also faced this in their schemas?

I had a look and noticed that we have only added `nullable` to properties, not to `geometry`. I will create a PR in Features.

m-mohr:
> The change by @tomkralidis is correct. Yours is not a valid OpenAPI schema.

Sorry, but I disagree. Why is this invalid OpenAPI? No type set -> all types allowed -> nullable adds null to the allowed types -> enum restricts to a single value (null).

m-mohr (changes requested):
The current approach doesn't work for null. Validating geometryGeoJSON.yaml against a null value will fail!

cportele:
> Why is this invalid OpenAPI? No type set -> all types allowed -> nullable adds null to the allowed types -> enum restricts to a single value (null).

`nullable` does not change the semantics in this case:

> nullable: A `true` value adds `"null"` to the allowed type specified by the `type` keyword, only if `type` is explicitly defined within the same Schema Object.

The other issue is that even if it would conform to OpenAPI 3.0.3, it would allow **any** value (e.g., `geometry: 42`). 

For more discussion, see the following proposal that was adopted for v3.0.3: https://github.com/OAI/OpenAPI-Specification/blob/main/proposals/2019-10-31-Clarify-Nullable.md.

On the other hand, you state that ...

> The current approach doesn't work for null. Validating geometryGeoJSON.yaml against a null value will fail!

Why would that be the case. Because a `null` value will violate the `required` constraints? The page linked above clarifies that these do not apply in the `null` case.

`nullable` was really a bad design and it is good that they got rid of it again.

tomkralidis:
FYI geometry null handling updated accordingly.

PR #265, opened 2023-05-24
update time null to align with geometry

tomkralidis:
cc @m-mohr


m-mohr:
If this is the only reference to time.yaml, the nullable works better than the enum and I'd not merge this. 
The nullable didn't work in the other example due to the usage of oneOf and multiple nullable subschemas, but that's not the case here.

pvretano (approved):

PR #266, opened 2023-05-25
Revert "update time null to align with geometry"

pvretano:


pvretano:
Oops, I just saw @m-mohr comment and I think I agree in this case.  @tomkralidis please review and either merge this PR to undo the previous one OR delete this PR.
This is @m-mohr comment: https://github.com/opengeospatial/ogcapi-records/pull/265#issuecomment-1561937878

tomkralidis:
I will someone testjng tonight with check-jsonschema and report back.

tomkralidis:
FYI I've merged the above until we can verify further.

@m-mohr @pvretano @kalxas I've put forth [this gist](https://gist.github.com/tomkralidis/cc674035706640d93a1f9527f18cb91a) for review/comment.  Note that I am using the schema in a non-OpenAPI context here in support of record validation.

PR #289, opened 2023-06-06
Patch formatting problems and fix other minor inconsistencies.

### LINKED ISSUE #278: Req. 9/10: various smaller issues
1) Req. 9A talks about 

> $.collections[*].itemType

but at that point the response body has not been defined to it's confusion to read this statement. It is defined in Req. 10 through the link to collections.yaml.

---

2) The other small inconsistency is that Req. 9A requires `itemType` to be `record`, but the schema in Req. 10 doesn't actually enforce this.

I think the following should be added to the properties:
```yaml
  itemType:
    enum:
      - record
```

--- 

3) Lastly, req. 10 is confusing because it talks about the response body in general and the individual collection encoding in the same sentence:

> The body of the response SHALL contain a description of the catalogue based upon the following schema fragment which extends the [collections.yaml](http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/collections.yaml) schema defined in the [OGC API - Features - Part 1: Core](http://docs.ogc.org/is/17-069r3/17-069r3.html) standard:

The collections.yaml (with s) finally defines the response body, but the following yaml provided in Req. 10 only extends/overrides the collection.yaml (without s). It should be made more clear that this is an extension to the collection.yaml that is linked to from the collections.yaml.

### LINKED ISSUE #279: Wrong formatting in Req. 11 B / 19 B
The formatting seems bugged:

grafik

The same applies in 19 B.

pvretano:
Closes #278 
Closes #279


core/standard/requirements/records-api/REQ_record-collections-response.adoc
```diff
@@ -2,6 +2,6 @@
 [width="90%",cols="2,6a"]
 |===
 ^|*Requirement {counter:req-id}* |*/req/records-api/record-collections-response*
-^|A |Only collections where the value of the `itemType` property (JSONPath: `$.collections[*].itemType`) is `record` SHALL be considered record collections or catalogues.
+^|A |For an operation that select a list of collection objects, only collections where the value of the `itemType` property (JSONPath: `$.collections[*].itemType`) is `record` SHALL be considered record collections or catalogues.
```
m-mohr, line 5: Can you switch the order of A and B? Makes it easier to read.

kalxas (approved):

PR #290, opened 2023-06-06
Update templated links to support both aware and unaware clients.

### LINKED ISSUE #275: Templated links conflict with OGC API - Common?!
In OGC API - Records I recently found the chapters about "Templated links with variables":
- http://docs.ogc.org/DRAFTS/20-004.html#sc_templated_links_with_variables
- http://docs.ogc.org/DRAFTS/20-004.html#_templated_links_with_variables

I think the addition of templates links with variables is not correctly extending OGC API - Common and is actually in conflict with it (also with STAC).

If I understand [Common, ch. 6.3](https://docs.ogc.org/is/19-072/19-072.html#_11b9b4f7-42fc-413a-b63a-e7fb060b5e4b) correctly, the href must be a valid IRI/URI according to RFC 8288.

This implies to me that with a client, all `href`s should be resolvable and if I render them e.g. in a HTML document as `href` parameter in an `a` tag, user can click the link without issues and end up at a "valdid" page.

The way Records adds templated links, I think this would break clients that are not aware of templated links. If I understand correctly, that `templated` must be set to `true` and then `href` is containing variables. As such the `href` won't resolve without specifying the values for the variables and my HTML example with the `a` tag won't work anymore.

A simplified example from Records:
```json
{
  "rel": "item",
  "type": "image/png",
  "title": "World Ozone and Ultraviolet Radiation Data Centre (WOUDC) stations",
  "href": "https://geo.woudc.org/ows?service=WMS&version=1.3.0&request=GetMap&crs={crs}",
  "templated": true,
  "variables": {
    "crs": {
      "description": "...",
      "type": "string",
      "enum": [
        "EPSG:4326",
        "EPSG:3857"
      ]
    }
  }
}
```

The href as such doesn't work in a naive (non-template-aware) client. The `templated` property basically weakens the href requirements, which is basically against the Liskov Substitution Principle.

I think Records has to extend Common in a different way. It must provide a default href that works (with defaults for the variables) and then add variables on top for clients that are aware of this functionality.
It could look like this for example:
```json
{
  "rel": "item",
  "type": "image/png",
  "title": "World Ozone and Ultraviolet Radiation Data Centre (WOUDC) stations",
  "href": "https://geo.woudc.org/ows?service=WMS&version=1.3.0&request=GetMap&crs=EPSG:4326",
  "templated": "https://geo.woudc.org/ows?service=WMS&version=1.3.0&request=GetMap&crs={crs}",
  "variables": {
    "crs": {
      "description": "...",
      "type": "string",
      "enum": [
        "EPSG:4326",
        "EPSG:3857"
      ],
     "default": "EPSG:4326"
    }
  }
}
```

In addition to all this, are templates variables really needed in the core?

pvretano:
Closes #275 

This is just a draft pending discussion in the SWG about templated links.

The gist of the PR is that client who know nothing about templated links should still be able to resolve the link.  

This is accomplished by converting the `templated` member from a boolean to a string that supplies the text of the parameterized URL.  If it is present then the `href` member simply contains an example or representative URL created by picking valid substitution values from the dictionary of substitution variables.

Aware clients can ignore the `href` value and generate a new resolvable URL using the dictionary of substitution variables.

Unaware clients can ignore all the templated stuff and simply use the value of the `href` member.


m-mohr:
I'd prefer to use variables instead of varBase (or combine them). The additional HTTP request for links is not a good ideas in web UIs.

cportele (commented):
If we decide to change the approach, then I think we should follow the approach of the proposed ID (https://ietf-wg-httpapi.github.io/link-template/draft-ietf-httpapi-link-template.html). That is, Link and LinkTemplate should be separate objects (like they are or will be separate headers). Conflating the two is really what the issue is about that triggered this PR. We would need an additional member in addition to the `links` array; e.g., `templates` for an array of link templates.

In any case, making `href` optional should not be an option and it won't help clients that expect a link (and cannot handle a URI template). `href` and `rel` are should continue to be manadatory in a link. I also think that in most cases, a sample URL will not be helpful to clients. How does a link to a single tile help any client when this is really about a xyz link template to a tile set?

`varBase` is also not the URI of some document. It is a base URI to which the name of the variable is added. However, the Link-Template Header is still a draft, so there may be changes as the draft progresses.

m-mohr:
@cportele That's also a good option, I think. My main point is really just that templated links should not interfere with the links in a way that it breaks if clients (or servers) are not aware of it.

cportele:
@pvretano - You have added the link templates in the `links` array, which does not resolve the issue. Clients that expect only links in that array will run into issues, if the encounter an object without a `href` member.

Also, `varBase` is not consistent with the Link-Template ID (see my earlier comment).

m-mohr:
Interesting, I understood this a bit differently. I had thought we keep the new object out of the links and ideally create a new property. 

Simplified Example:
```
"links": [
  { "href": "example.json", "rel": "example" }
],
"linkTemplates": [
  {
    "href": "example.json?q={var}",
    "rel": "example",
    "variables": {
      "var": { "type": "string" }
    }
  }
]
```

m-mohr:
@pvretano No, because in code they would expect that the href is there, because it is defined in several specs that href is required and the array consists of Link Objects only.

I know people do stuff like (pseudo code):
````
link = links.find(l =>l.rel == 'abc')
http.get(link.href)
````

pvretano:
@m-mohr, @cportele please review ...

kalxas (approved):

pvretano:
05-JUL-2023:  Need to a requirement that says that the information returned by resolving a URI composed of the `varBase` and the variable name SHALL be JSON-Schema.  @pvretano will make this addition and then merge this PR.

m-mohr:
We've also had a larger discussion during FOSS4G about this. General consensus was that templates links should be separate unless we add templated to Common, Features and STAC. Otherwise, variables + varBase look good enough.

(stac_folks didn't like the camelCase, but we run into these issues anyway with numberReturned etc, so ignore it for now.)

pvretano:
10-JUL-2023:  OGC API Tiles, an adopted specification, already uses templated links without separating them into two sections (links, linkTemplates).  Features does the same.  In both cases they follow the IETF standard and use `varbase` for getting information about the variable.  So, if we don't merge this PR and remove the ability to specify the variable descriptions in-line then Tiles, Features and Records would be aligned.  @pvretano and @kalxas think this is probably the best approach for Records.
We probably need a broader discussion in OGC about whether we need to have separate sections for links and linkTemplates but so far, at least for Tile and Features, that does not seem to be a problem.  @pvretano will ask @ghobona about the best forum for that discussion.

cportele:
@m-mohr - It is currently at "Unresolved directive in clause_7_core.adoc - include::recomendations/core/REC_fc-md-item-link-templates.adoc[]" ;) I just pushed a fix, it should be included in the Editor's draft tomorrow. 

The direct link to the recommendation is here: https://github.com/opengeospatial/ogcapi-features/blob/master/core/standard/recomendations/core/REC_fc-md-item-link-templates.adoc

pvretano:
Converting to drafts because of [this](https://github.com/opengeospatial/ogcapi-records/pull/290#issuecomment-1629096679) discussion.

m-mohr:
@cportele Thanks for the pointer. So this is going to be Features 2.0? Strictly speaking this is a breaking change, I'd say:

1. In theory, you are now allowed to specify variables for e.g. next or self links, which are not necessarily dereferencable with variables.
2. The html conformance class says all links should be available in <a> tags, but this doesn't allow variables (afaik)
3. Links should be provided as HTTP Links, but RFC8288 (which Features references to), doesn't allow templates (afaik)

To me this looks like OGC made a mistake in Tiles, but now instead of fixing it (that's certainly possible in a non-breaking way in Tiles), it repeats the mistake again.

cportele:
@m-mohr - That's why we said that we need to have a broader discussion across the OGC API groups. I think that the best approach is to introduce a separate array `linkTemplates` for the link templates as in this PR, but this should not just be a Features or Records decision / approach given that Tiles was the first to use link templates and decided to use a different approach.

m-mohr:
@cportele Thanks, I missed that. Where is this discussion taking place?

tomkralidis (approved):

kalxas (approved):

pvretano:
13-NOV-2023: @pvretano will resolve the merge conflicts.  The consensus of the SWG is that we remove the ability to specify variable definition in-line and simply stick with the RFC which only defines varbase.  @pvretano will check with @tomkralidis if removing the inline option is OK and if it is @pvretano will merge.  If not, will discuss at the next SWG meeting.

tomkralidis:
+1/fine with me to removing inline option.

m-mohr:
Not overly happy with removing inline as it may lead to quite a number of additional requests. Resolving a link will not be self-contained, which I feel adds quite a bit of complexity. What were arguments for removing it?

cportele (changes requested):
Many cases still have "href", not "link-template", otherwise it looks good.

jerstlouis:
@pvretano @cportele Since this affects all the SWGs, could we adopt someting else than `link-template` because that does not follow the decision to avoid hyphens `-` in JSON property names. The _Common_ guidance was to use camelCase for JSON property names, kebab-case for query parameters, although the final decision was not fully documented in [issue](https://github.com/opengeospatial/ogcapi-common/issues/224).

Potentially something shorther like `tref` or just `template` could be considered too, since this is already inside a _template_` object? In DGGS we used `tref` so far:

https://github.com/opengeospatial/ogcapi-discrete-global-grid-systems/commit/59b2a1c9972eb5e19d06cba53c0313bf3e724a1d

PR #291, opened 2023-06-07
Remove the redundant properties member.

### LINKED ISSUE #274: Additional properties in collections?
Looking at the record collection schema ( http://docs.ogc.org/DRAFTS/20-004.html#record-collection-schema ), I'm confused about how to add additional properties.

The "Table of collection properties" (Table 11) lists a `properties` field for "Additional catalogue-specific properties".

On the other hand, Permission 2 says:
- A collection object MAY contain any number of additional properties not listed in Table 11. The meaning of these additional properties is not defined in this standard.

I assume the "collection object" in Permission 2 is the (JSON) top-level object that also contains the `type`, `id`, `title`, `license`, ... properties.

Where to place additional properties? Top-level? In properties? STAC places them in the top-level, both fields that are in core STAC (e.g. assets), but also extensions (e.g. `sci:doi` from the scientific extension). Thus my preference it top-level, of course ;-)

pvretano:
Closes #274

### DIFF
--- a/core/openapi/schemas/catalogue.yaml
+++ b/core/openapi/schemas/catalogue.yaml
@@ -1,6 +1,7 @@
 ---
 allOf:
-  $ref: "http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/collection.yaml"
+  $ref: "http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/\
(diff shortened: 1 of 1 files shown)


m-mohr:
I think you also need to update table 11, i.e. remove the last row.

pvretano:
Yup ... done.

PR #296, opened 2023-07-05
clarify name/organization (#294)

### LINKED ISSUE #294: clarify contact name/organization
The [current contact schema](https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/contact.yaml) describes contact `name` / `organization` as follows:

```yaml
  name:
    type: string
    description: |-
      The name of the organization or the individual.
  organization:
    type: string
    description:
      Organization/affiliation of the individual/responsible person. In
      case of an organization, the name property should be used and this
      property is not to be used.
```

`name` is a required property.

Is implementing `name` or `organization`, then, domain/activity specific?  For example, consider a catalogue implementing facets on contact organizations.  Would the catalogue have to check if `organization` exists?  If not, then use `name` and assume `name` is an organization?

Should we consider a contact to require ~ONE OF~ ANY OF `name` or `organization`, with clear definitions of each?

cc @pvretano @kalxas @m-mohr @pvgenuchten

tomkralidis:
Fixes #294.  As well, renames `positionName` to `position`; the former appears to be carried over from ISO 19115 (`positionName`, `organizationName`, etc.), so changed for consistency here.

### DIFF
--- a/core/openapi/schemas/contact.yaml
+++ b/core/openapi/schemas/contact.yaml
@@ -3,18 +3,21 @@ type: object
 description: |-
   Identification of, and means of communication with, person responsible
   for the resource.
-required:
-  - name
+anyOf:
+  - required:
+    - name
+  - required:
+    - organization
 properties:
   identifier:
     type: string
     description: |-
-      A value uniquely identifying a contact (individual or organization)
+      A value uniquely identifying a contact.
   name:
     type: string
     description: |-
-      The name of the organization or the individual.
-  positionName:
+      The name of the responsible person.
+  position:
     type: string
     description:
       The name of the role or position of the responsible person taken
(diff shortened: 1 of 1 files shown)


core/openapi/schemas/contact.yaml
```diff
@@ -3,28 +3,31 @@ type: object
 description: |-
   Identification of, and means of communication with, person responsible
   for the resource.
-required:
-  - name
+anyOf:
+  - allOf:
```
m-mohr, line 7: You don't need the allOf in both branches :-)

tomkralidis: Confirmed this continues to work as expected.  PR updated.

pvretano (approved):

PR #297, opened 2023-07-06
add recommendation for contact country as ISO 3166 alpha-3 (#261)

### LINKED ISSUE #261: ISO code for country?
I'm implementing the contacts field right now and the country field in the address would be much more helpful (e.g. for address formatting) if it was just an [ISO 3166](https://en.wikipedia.org/wiki/ISO_3166) (alpha-2?) code instead of a free-form text.

I'm wondering whether we should:
- require it?
- recommend it?
- just mention it?
- ignore it? ;-)

tomkralidis:
Fixes #261

### DIFF
--- a/core/openapi/schemas/contact.yaml
+++ b/core/openapi/schemas/contact.yaml
@@ -95,7 +95,7 @@ properties:
           description: ZIP or other postal code.
         country:
           type: string
-          description: Country of the physical address.  ISO 3166-1 alpha-3 is recommended.
+          description: Country of the physical address.  ISO 3166-1 is recommended.
         roles:
           description:
             The type of address (e.g. office, home, etc.).
--- a/core/openapi/schemas/contact.yaml
+++ b/core/openapi/schemas/contact.yaml
@@ -95,7 +95,7 @@ properties:
           description: ZIP or other postal code.
         country:
           type: string
-          description: Country of the physical address.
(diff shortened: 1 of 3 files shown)


m-mohr:
Are there any statistics for alpha-2 vs. alpha-3? I feel like I see alpha-2 much more often, but might be biased...

Otherwise, looks good.

tomkralidis:
alpha-2 is probably the most popular.  alpha-3 is used in a few geospatial metadata standards.

How about relaxing to just "ISO 3166" (which would then allow for alpha-2, alpha-3, or numeric)? Some tooling ([Python](https://pypi.org/project/iso3166/), [JavaScript](https://www.npmjs.com/package/iso-3166)) have support for these interchangeably.

m-mohr:
I don't feel strongly either way, so that works for me.

tomkralidis:
PR updated.

cnreediii:
The internet RFC's that are part of the standards framework for the NG-911 deployment use alpha-2. There is a core concept called a location object. There is a geodetic (coordinates) and a civic (address) LO. The civic LO uses alpha-2 so all NG-911 deployments in the US, Canada and other nations (Europe) will use alpha-2. The civic LO standard is referenced as normative in numerous other internet standards including DHCP.

kalxas (approved):

pvretano (approved):

PR #310, opened 2023-10-10
STAC -> OARec crosswalk (plus some minor editorial fixes).

### LINKED ISSUE #306: Add informative text cross-walking OARec concepts with STAC concepts.
It was suggested, at the Sept 2023 members' meeting, to add information text to the specification cross walking OGC API Records concepts with STAC concepts.

pvretano:
Closes #306


kalxas (approved):

core/standard/annex_stac.adoc
```diff
@@ -0,0 +1,237 @@
+[appendix]
```
m-mohr, line 1: Would it make more sense to just link to the original document? I don't really want to maintain this information in two places and I think this is more a living document which evolves over time while we further align and evolve (e.g. STAC 1.1). Having it as part of a published standard doesn't feel right.

pvretano: 16-OCT-2023 SWG Meeting: In general the SWG believes that it would be better to link out to this information.  However, it was copied into the standard because we are not sure about the persistence of the source material.  @pvretano will ping @m-mohr about that and update the document accordingly. @kalxas mentioned that the source material is in the [STAC UTILS](https://github.com/stac-utils) repo and might be better if it was closer to the [STAC specification](https://github.com/radiantearth/stac-spec) repo.

m-mohr: I have access to stac-spec, but we try to keep non-normative separate, we'll even try move the best practices out. So it's unlikely that the comparison document will move to stac-spec. Also, this was never reviewed by the STAC PSC, it was just me writing this as part of another project. What's the concern here? That the status is not stable (e.g. we may fix issues over time)? Or that it just goes offline at some point?

cnreediii:
@m-mohr Thanks. Peter et.al. perhaps we (OGC) move the crosswalk document to the OGC Git repo with a note about the source of the document? And then link to the OGC location?

m-mohr:
FYI: The document will be part of the Testbed 19 ER, so you could also link to that. The link will be stable but not be updated over time.

It depends on what you want though. I'll keep the document in stac-crosswalks updated for a bit, but it's unlikely that it goes away. So even if we decide to move the repo out of stac-utils, GitHub is good at redirecting the links of repos that got moved around.

pvretano:
13-NOV-2023: SWG feels referencing is better than copying.  So, this PR can be merged.

PR #314, opened 2023-10-10
Enhance the description and capabilities (slightly) of the q parameter.

### LINKED ISSUE #295: q: Search terms delimited by comma?!
Porting over the discussion from https://github.com/opengeospatial/ogcapi-records/pull/273#discussion_r1218620027_

@m-mohr wrote:
> How are search terms defined and generated from the q parameter? Is it basically: `terms = q.split(' ')` ? So is space the only separator and everything in-between spaces (or the beginning/end of the string) is a term?

@pvretano wrote:
> The definition of the parameter is:
> 
> ```
> ---
> name: q
> in: query
> required: false
> schema:
>   type: array
>   items:
>     type: string
> explode: false
> style: form
> ```
> 
> so the search terms are comma-separated and so the **comma** delimits the terms.

@m-mohr wrote:
> @pvretano Good point, on the other hand that's not what users expect, I think. Users type free text phrases with spaces into Google, I've never seen anyone type commas. This would eventually mean that you are not able to simply pass through the text field value, but instead you have to do something like `searchTerm.replace(' ', ',')` beforehand. That just doesn't feel right to me, but maybe it's just me?

pvretano:
Closes #295

### DIFF
--- a/core/standard/clause_7_building_blocks.adoc
+++ b/core/standard/clause_7_building_blocks.adoc
@@ -492,9 +492,32 @@ include::requirements/record-core-query-parameters/REQ_query-param-limit.adoc[]
 [[core-query-parameters-q]]
 ===== Parameter q
 
-include::requirements/record-core-query-parameters/REQ_query-param-q-definition.adoc[]
+The `q` parameter is meant to provide and rudimentary, easy to implement keyword search capability across one or more text fields in a record.  The specific text fields in a record that are to be searched is not mandated in this standard.
 
 include::recommendations/record-core-query-parameters/REC_param-q.adoc[]
+
+The `q` parameter is a comma-separated list of search terms.  If any of these terms appear in a searched text field of a record then that record can be included in the result set (logical OR).  If any search term specified using the `q` parameter contains one or more spaces, this implied that all of the specified search terms must appear in a searched text field of a record before that record can be included in the result set (logical AND).
+
+For example, the following `q` parameter specification:
+
+----
+q=ocean,climate   change,desalination
+----
+
+represents the query predicate:
+
+----
+(anytext CONTAINS 'ocean') OR
+(anytext CONTAINS 'climate' AND anytext CONTAINS 'change') OR
+(anytext CONTAINS 'desalination')`
+----
+
+NOTE:
+
+* The term `anytext` is meant to represent the set of text fields that are searched by the operation.
+* The `CONTAINS` operator used above is just an example of a operator that searches a text field(s) for a specific search term.  How this operator is implemented and what its named in the system backing a catalog implementation is are beyond the scope of this standard.
+
+include::requirements/record-core-query-parameters/REQ_query-param-q-definition.adoc[]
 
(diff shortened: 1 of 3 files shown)


core/standard/clause_7_building_blocks.adoc
```diff
@@ -492,10 +492,34 @@ include::requirements/record-core-query-parameters/REQ_query-param-limit.adoc[]
 [[core-query-parameters-q]]
 ===== Parameter q
 
-include::requirements/record-core-query-parameters/REQ_query-param-q-definition.adoc[]
+The `q` parameter is meant to provide and rudimentary, easy to implement keyword search capability across one or more text fields in a record.  The specific text fields in a record that are to be searched is not mandated in this standard.
 
 include::recommendations/record-core-query-parameters/REC_param-q.adoc[]
 
+The `q` parameter is a comma-separated list of search terms.  If any of these terms appear in a searched text field of a record then that record can be included in the result set (logical OR).  If any search term specified using the `q` parameter contains one or more spaces, this implied that all of the specified search terms must appear in a searched text field of a record before that record can be included in the result set (logical AND).
```
cportele, line 499: Another, more natural interpretation of spaces in a search term would be that the search term including the spaces would match. That is, in the example `anytext CONTAINS 'climate change'`.

Both options are useful, but I would use the more intuitive option and leave the logical expressions to CQL2. That said, if there is a preference for the AND interpretation, that is OK for me, too.

cportele (approved):

kalxas (approved):

core/standard/clause_7_building_blocks.adoc
cportele: @pvretano Not that I can "see" the four spaces in the markdown rendering 😉 , but I agree that it should be any whitespace separating the words.

m-mohr: I agree. I'd expect that if I search for `q=climate change` that I'd get all results that contain "climate change", but not the results that contain "change climate"

pvretano:
30-OCT-2023: q parameter PR will update to reflect the logical that comma-separated terms are OR'ed and terms with white spaces are matched as a single term but any number of white spaces separating the terms. @pvretano to update PR to reflect this logic.

kalxas (approved):

tomkralidis (approved):

pvretano:
13-NOV-2023: @pvretano to update the PR to (a) remove the use of "logical AND" since what is described with relation to white space separated search terms is not really and logical AND and (b) remove item "D" from the "q" definition.  After those changes have been made, this PR can be merged.

m-mohr:
I didn't get to checking whether this has any implications to the STAC extension, sorry. 
It might need some additional clarifications regarding the space I assume: https://github.com/cedadev/stac-freetext-search#basic

kalxas (approved):

pvretano:
SWG meeting 11-DEC-2023: The concensus of the SWG is to merge with the understanding that the wording in the requirements could be simplified.  We will merge and then await feedback from reviewers on this point.

PR #322, opened 2023-12-29
Update templated links to support both aware and unaware clients.

### LINKED ISSUE #275: Templated links conflict with OGC API - Common?!

pvretano:
Closes #275

The gist of the PR is that client who know nothing about templated links should still be able to resolve the link.

This is accomplished by converting the templated member from a boolean to a string that supplies the text of the parameterized URL. If it is present then the href member simply contains an example or representative URL created by picking valid substitution values from the dictionary of substitution variables.

Aware clients can ignore the href value and generate a new resolvable URL using the dictionary of substitution variables.

Unaware clients can ignore all the templated stuff and simply use the value of the href member.

### DIFF
--- a/core/examples/json/associations-simple-and-uri-template.json
+++ b/core/examples/json/associations-simple-and-uri-template.json
@@ -5,19 +5,15 @@
       "type": "text/html",
       "title": "WOUDC - Data - Station List",
       "href": "https://woudc.org/data/stations"
-    },
+    }
+  ],
+  "linkTemplates": [
     {
       "rel": "item",
       "type": "image/png",
       "title": "World Ozone and Ultraviolet Radiation Data Centre (WOUDC) stations",
-      "href": "https://geo.woudc.org/ows?service=WMS&version=1.3.0&request=GetMap&crs={crs}&bbox={bbox}&layers=stations&width={width}&height={height}&format=image/png",
-      "templated": true,
+      "link-template": "https://geo.woudc.org/ows?service=WMS&version=1.3.0&request=GetMap&crs={crs}&bbox={bbox}&layers=stations&width={width}&height={height}&format=image/png",
       "variables": {
-        "crs": {
-          "description": "...",
-          "type": "string",
-          "enum": [ "EPSG:4326", "EPSG:3857" ]
-        },
         "bbox": {
           "description": "...",
           "type": "array",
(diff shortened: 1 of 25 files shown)


pvretano:
The merge conflicts in PR #290 were so convoluted that is was simpler to just create a new PR based on the latest master!

PR #329, opened 2023-12-30
Patch OpenAPI example ...

### LINKED ISSUE #309: Validate all examples
See #308 for an example of need for this.

Set up automated validation of examples to avoid this problem. 

Contact OGC COSI staff for assistance if required. 

OGC Building Blocks templates use a validation pipeline available from https://github.com/opengeospatial/bblocks-postprocess

This could be directly applied or adapted to use the existing repository, or a new OGC building blocks repository set up as a wrapper for all normative and informative content to perform validation and provide an opportunity to undertake further automation such as exercising mappings to other schemas.

pvretano:
Patch OpenAPI building blocks file and example.  Convert all instances of "catalogue" to "catalog".

Closes #309


core/openapi/schemas/schema.yaml
tomkralidis, line 1: Do we need the OpenAPI schema itself here?

tomkralidis (changes requested):

core/openapi/schemas/schema.yaml
pvretano: Yup ... the list of sortables is a list of schema fragments that describe each sortable.  I suppose we could just say `"type": "object"` and simply say that the object has to be a schema fragment.

tomkralidis (approved):

kalxas (approved):

PR #331, opened 2024-01-03
update schema refs

tomkralidis:


core/openapi/ogcapi-records-1.yaml
```diff
@@ -237,7 +237,7 @@ components:
                 default: 'http://www.opengis.net/def/uom/ISO-8601/0/Gregorian'
       landingPage:
         allOf:
-          - $ref: "#/components/schemas/common/link.yaml"
```
pvretano, line 240: This should be `#/components/schemas/common/link`.  It's supposed to be an internal reference.

tomkralidis: When do we use internal references vs. `$ref`?  I see both in this file.

core/openapi/ogcapi-records-1.yaml
```diff
@@ -237,7 +237,7 @@ components:
...
+          - $ref: schemas/common/link.yaml
```
pvretano, line 240: Should be: "#/components/schemas/common/link"

pvretano (approved):
@tomkralidis, look good.

PR #338, opened 2024-01-13
Schema reorg experiment #3.

pvretano:
This PR is part of an internal exercise among the active specification editors (@pvretano, @tomkralidis, @kalxas ) to determine the best way to lay out the openapi directory.  The other PRs that are part of this exercise are PR #336 and PR #337.

The editors  will evaluate the PRs and select one to merge into the specification.  The others will be closed.  Since Records depends on Features for its API, the most likely candidate is PR #337 since it uses the same openapi layout/approach that is used in Features.


pvretano:
Archiving because PR #337 was merged.

Issue #349, opened 2024-03-25
incompatibility with STAC - time property


rob-metalinkage:
records makes `time` required,  

using the OGC building block register tooling for testing STAC is a valid extension [1]  identifies that all STAC examples fail when tested against record schema.

see https://ogcincubator.github.io/bblocks-stac/build/tests/report.html

either Records or STAC needs to change, and it appears Records would be easiest.  ( a profile of records with mandatory "time" may be a solution if there is a requirement.)

[1]  https://ogcincubator.github.io/bblocks-stac/

PS:  BBlocks for Features and Records is in progress to test all examples in the base specs.

m-mohr:
See also https://github.com/opengeospatial/ogc-feat-geo-json/issues/69

rob-metalinkage:
can be closed by https://github.com/opengeospatial/ogcapi-records/pull/353

PR #353, opened 2024-04-14
Make the "time" member optional to align with STAC.

pvretano:
In STAC, the time member is optional so, in order to promote more alignment between the two standards (since we want STAC to be a profile of Records), I've made the changes to make the top-level `time` member optional.  I did, however, add a recommendation that says that the `time` member shoud be included in the record.

### DIFF
--- a/core/standard/clause_7_building_blocks.adoc
+++ b/core/standard/clause_7_building_blocks.adoc
@@ -51,9 +51,9 @@ Other encodings are allowed but are not described in this document.
 |===
 |Property |Requirement |Description |GeoJSON key
 |geometry |**required** |A geometry associated with the resource that is used for discovery.  Can be null if there is no associated geometry. |`geometry`
-|time |**required** |The temporal extent of the resource. Can be null if there is no associated temporal extent. |`time`
 |type |**required** |The nature or genre of the resource. |`properties.type`
 |title |**required** |A human-readable name given to the resource. |`properties.title`
+|time |optional |The temporal extent of the resource. Can be null if there is no associated temporal extent. |`time`
 |description |optional |A free-text description of the resource. |`properties.description`
 |keywords |optional |A list of free-form keywords or tags associated with the resource. |`properties.keywords`
 |language |optional |The language used for textual values (i.e. titles, descriptions, etc.) of this record. |`properties.language`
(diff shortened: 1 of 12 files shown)


core/standard/clause_7_building_blocks.adoc
```diff
@@ -51,9 +51,9 @@ Other encodings are allowed but are not described in this document.
 |===
 |Property |Requirement |Description |GeoJSON key
 |geometry |**required** |A geometry associated with the resource that is used for discovery.  Can be null if there is no associated geometry. |`geometry`
-|time |**required** |The temporal extent of the resource. Can be null if there is no associated temporal extent. |`time`
 |type |**required** |The nature or genre of the resource. |`properties.type`
 |title |**required** |A human-readable name given to the resource. |`properties.title`
+|time |optional |The temporal extent of the resource. Can be null if there is no associated temporal extent. |`time`
```
tomkralidis, line 56: If `time` is now optional, do we care about specifying null rather than not specifying `time` at all?  Suggest the description be updated to only `The temporal extent of the resource.`

core/standard/clause_7_building_blocks.adoc
```diff
@@ -99,7 +99,11 @@ Records can have multiple properties that describe the temporal characteristics
...
+This Standard does not place any constraints on the number of additional temporal properties that may be added to a record to characterize temporal aspects of the resource being described by a record.  This can make it difficult for a naive client to recognize and utilize temporal information stored in the record.  For this reason, this Standard includes the `time` property in a record.  
+
+include::recommendations/record-core/REC_time.adoc[]
+
+This `time` member is set by the publisher of the record and describes characteristic temporal information (an instant or an interval) associated with the resource described by a record.   This `time` member can thus be used by naive clients without the need to inspect or understand the schema of an extended record.  If included in the record, the value of the `time` member is either `null` (no temporal information) or an object encoding a time instant or interval.
```
tomkralidis, line 106: `If included in the record, the value of the `time` member is an object encoding a time instant or interval.`

core/standard/recommendations/record-core/REC_time.adoc
```diff
@@ -0,0 +1,7 @@
+[[rec_record-core_time]]
+[width="90%",cols="2,6a"]
+|===
+^|*Recommendation {counter:rec-id}* |*/rec/core/record-contact*
+^|A |A record SHOULD include the top-level `time` member that indicates some characteristic temporal aspect of the resource that the record describes.
```
tomkralidis, line 5: `A record SHOULD include the top-level `time` member that indicates the temporal aspect of the resource that the record describes`

core/standard/recommendations/record-core/REC_time.adoc
```diff
@@ -0,0 +1,7 @@
...
+^|B |If such chracteristic temporal information about the resource is not available then the `time` member SHOULD, even though it is optional, still be included in the record and its values SHOULD be set to `null`.
```
tomkralidis, line 6: Here as well, why null vs not encoding at all?

tomkralidis (changes requested):
See comments.  While we are here, does this have any impact on FG-JSON alignment (which requires `time`)?

core/standard/clause_7_building_blocks.adoc
pvretano: We do care.  Even though we are making time optional to be compatible with STAC, we should  try not to impede eventual comparability with JSON-FG.  Allowing time to be present and null allows comparability with the JSON-FG time element.

tomkralidis: We were already directly compatible with JSON-FG.  So I guess this is a middle ground. Should JSON-FG consider making a similar change? This is outside the scope of this PR, of course.

rob-metalinkage:
It would seem necessary to formalise optional FG JSON support as a named profile (conformance class). We could also define BP profiles including constraints implementing recommendations.  I'll build some STAC building blocks based on such recommendations, allowing people to transparently declare the behaviour.

core/standard/clause_7_building_blocks.adoc
pvretano: @tomkralidis if I am remembering correctly, the last discussion was had about this at the London Code (2023) sprint was that we would NOT add a JSON-FG conformance class to the standard.  So, if we are compatible with JSON-FG (which I actually don't think we are ... but I could be wrong), it would only be by conicidence.  So, yes, this is a middle ground.  As for the JSON-FG SWG considering making the time member optional, that is something for them to consider.

core/standard/clause_7_building_blocks.adoc
pvretano: I updated the schemas to allow time to be `null` or an object.

core/standard/recommendations/record-core/REC_time.adoc
pvretano: Done.

core/standard/recommendations/record-core/REC_time.adoc
pvretano: See [here](https://github.com/opengeospatial/ogcapi-records/pull/353#discussion_r1564878747).

rob-metalinkage:
closes https://github.com/opengeospatial/ogcapi-records/issues/349

tomkralidis (approved):

kalxas (approved):

PR #358, opened 2024-05-01
Exploration PR to statisfy issue-300.

### LINKED ISSUE #300: Local Resource Catalogs and req. class organization
Following up on past suggestions via e-mail, and based on previous discussion and clarification from @pvretano, if I understand correctly the intent of [Section 12 - _Local Resource Catalogs_](https://docs.ogc.org/DRAFTS/20-004.html#clause-local-resources-catalogue) would allow to catalog OGC API collections, whether local or from other services, as local collections at `/collections/{collectionId}` (instead of as items at `/collections/{collectionId}/items/{itemId}`), so as to enable the concept of a super-API that could gather several data collections under a single API (which could still use remote-links to link to the access mechanisms like `/coverage`, `/tiles`, `/items` etc.). In addition, this would enable functionality to filter a large set of collections directly at `/collections` (also related to potential _Common - Part 2_ req classes for collection filtering), instead of requiring a special _catalog_ collection that is metadata mixed up with all the other collection which are actual data.

Accessing `/collections` as a catalog is particularly interesting for a generic OGC API clients (e.g., the GDAL OGCAPI driver), as it does not require the implementation of a dedicated _Records_ client and does not involve any additional step for the user. Furthermore, local processing resource can be integrated into the same API, and data integration/access mechanisms like DGGS and _Processes - Part 3: Workflows and Chaining_ makes for a very interesting API and system of systems.

If this could indeed be possible, I believe there might be some issue with the current requirements class layout.
In particular, the current stated dependencies of _Local Resources Catalogue_ on the _Record API_ and the _Record Collection_ might be problematic.

I made the following suggestion as a potential alternative requirements classes layout to address these issues:

**Requirement Classes defining resources**
   
- Core (concept of a record and core properties) -- should they really be called "core queryables" since they can also be used in "crawlable catalogues" that cannot be queried?

**Requirement Classes defining origins**

- Records Collection (dependency on Features, with the records as a collection at `/collections/{collectionId}/items/{recordId}`)
- Local Resources Catalogue (records at `.../{resources}/{resourceId}`) where `{resources}` could be e.g., _Collections_ (`/collections/{collectionId}`) or _Processes_ (`/processes/{processId}`) or _EO Scenes_ (e.g.. `/collections/sentinel2-l2a/scenes/{sceneId}`). This would make it possible to apply Records advanced filtering capabilities at `/processes`, `/collections`, `/collections/sentinel2-l2a/scenes/`. It would be good to have a mechanism to associate specifically with a particulary type of local resources, like "Common - Part 2: Collections" or "Processes", perhaps as requirements classes of those other standards (we are planning to revive the _Common SWG_ at the September meeting, meeting in the OGC API track).

**Requirement Classes defining query parameters**

- Queryable Records (`q=`, `type=`, `externalId=` )
- (for` datetime` and `bbox`, _Common_ or _Features_ or some other spec would actually be defining them)
- Records Filtering (`filter=`, not necessarily CQL2, since you could have `filter-lang=` and you would explicitly declare conformance to CQL2 just like with _Features - Part 3_)
- Records Sorting (`sort=`)
- Records property selection (`properties=`) -- you may want to get lots of info back for several records, but only want some key properties that makes the results a lot smaller

The only requirements class dependency would be of everything else on "Core" (and _Records Collection_ on _OGC API - Features_).

Any end-point to a resource supporting OGC API - Records "Core" requirements class is automatically a "Records API" and a "Crawlable catalogue".
If that API implements Full Text Search, Records Filtering, or Records Sorting, or Records property selection, it is a kind of "Searchable Catalog".
A "Searchable Catalog" profile could specify that all of these need to be implemented to conform to it.

Would the SWG consider any of these suggestions?

Thank you!
cc. @tomkralidis @joanma747

pvretano:
After a long discussion with Jerome we determined that three changes might satisfy his requirements.  Those changes are:

1. In a record collection (aka a catalog) allow records to be encoded in-line or by reference.  This would allows a static catalog to be set up with paging.
2. Make the query parameters in the Local Resources Catalog requirements class optional like the sorting and advanced filtering query parameters.
3. Make the `geometry` member in the record optional in the building block but mandatory in the GeoJSON encoding of the record.  This is actually consistent with csw:Record where the geometry is optional.

This PR is an exploration to capture all these changes.  This allows us to keep the current organization but still satisfy the use cases that Jerome had in mind in relatation to issue #300.

Closes #300.


kalxas (approved):

core/examples/yaml/processSummaryLocalResource.yaml
tomkralidis, line 1: Should this file be named `recordSummaryLocalResource.yaml` ?

core/examples/yaml/processSummaryLocalResource.yaml
```diff
@@ -0,0 +1,10 @@
+---
+allOf:
+  - $ref: 'recordCommonProperties.yaml'
+  - $ref: 'https://schemas.opengis.net/ogcapi/processes/part1/1.0/openapi/schemas/processSummary.yaml'
```
tomkralidis, line 4: Is pointing to an OGC API - Processes schema intentional?

core/standard/clause_7_building_blocks.adoc
```diff
@@ -1070,7 +1090,13 @@ include::requirements/json/REQ_record-content.adoc[]
...
+[source,YAML]
+----
+include::../openapi/schemas/recordCommonProperties.yaml[]
+----
+
+NOTE: Referenced schemas can be found at https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/common/recordCommonProperties.yaml[recordCommonProperties.yaml], https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/time.yaml[time.yaml], https://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/geometryGeoJSON.yaml[geometryGeoJSON.yaml], https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/language.yaml[language.yaml], https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/theme.yaml[theme.yaml], https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/contact.yaml[contact.yaml], https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/license.yaml[license.yaml], https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/common/linkBase.yaml[common/linkBase.yaml].
```
tomkralidis, line 1099: This note might be better formatted as a bulleted list?

tomkralidis (changes requested):

core/examples/yaml/processSummaryLocalResource.yaml
pvretano: Yes, this is an example of extents the `/processes/{processId}` endpoint to make is a local resources endpoint.

core/examples/yaml/processSummaryLocalResource.yaml
pvretano: Nope.  This ia an example of extending the processSummary to make a local resources object that can be search like a catalog record.

core/standard/clause_7_building_blocks.adoc
pvretano: Fixed.

tomkralidis (approved):

kalxas (approved):

tomkralidis (approved):

PR #360, opened 2024-05-12
Make title optional everywhere and type mandatory only for the catalog.

### LINKED ISSUE #359: STAC incompatibilities
Further testing against STAC examples has highlighted another inconsistency (c.f. #349) 

https://github.com/opengeospatial/ogcapi-records/blob/master/core/openapi/schemas/recordGeoJSON.yaml#L35

```   required:
      - type
      - title
```

which is inconsistent with STAC schema and examples which do not require these.

pvretano:
Closes #359 
Does a lot of other clean and reorg too to make the spec a little more readable ... I think.


tomkralidis (approved):

kalxas (approved):

rob-metalinkage:
Thanks folks for quick response - I'll continue testing... hopefully one day you can incorporate automated testing of building blocks into the design process - happy to support this - will present to OGC API common in Montreal on the techniques and tools available.

PR #363, opened 2024-05-14
Address OAB review comments.

pvretano:
Address review comments from the OAB before RFC release.

### DIFF
--- a/core/examples/json/sortables.json
+++ b/core/examples/json/sortables.json
@@ -21,7 +21,7 @@
       },
       "relevance": {
          "title": "Relevance of the result",
-         "description": "This is a synthetic sortable that the server computes on the fly to assign a relevance to the result with respect to the query parameters specified in the original request",
+         "description": "This is a synthetic sortable that the server computes on the fly to assign a relevance score to each result with respect to the query parameters specified in the original request.  How exactly the relevance score is computed is up to the discretion of the server.",
          "type": "number",
          "minimum": 0,
          "maximum": 10
--- a/core/examples/json/sortables.json
+++ b/core/examples/json/sortables.json
@@ -19,6 +19,14 @@
          "description": "This is a synthetic sortable that the server computes on the fly based on the spatial extent of the resource.  Resources without spatial extent have zero area.",
          "type": "number"
       },
+      "relevance": {
+         "title": "Relevance of the result",
+         "description": "This is a synthetic sortable that the server computes on the fly to assign a relevance to the result with respect to the query parameters specified in the original request",
+         "type": "number",
+         "minimum": 0,
+         "maximum": 10
+      },
+      
       "extent": {
          "title": "Spatio-Temporal Extent",
          "description": "The spatio-temporal coverage of the resource.",
--- a/core/standard/clause_7_building_blocks.adoc
+++ b/core/standard/clause_7_building_blocks.adoc
@@ -1,12 +1,12 @@
 [[building-blocks]]
-== Building Blocks
+== Common Components
 
 === Overview
 
-This Building Blocks clause specifies the requirements classes that define building blocks that can be used to define various <<catalog-deployments,catalog deployments>>.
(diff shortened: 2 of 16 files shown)


core/standard/clause_7_building_blocks.adoc
```diff
@@ -909,6 +911,14 @@ sort order.
 
 include::requirements/sorting/REQ_defaultSortOrder-definition.adoc[]
 
+==== Sorting by relevance (informative)
+
+A common practice is to sort search results by relevance.
+
+This capability is accomodated in this Standard by using <<per_sorting_sortables-synthetic,synthetic>> sortables.  That is, sortables that are derived or synthesized by the server.
```
tomkralidis, line 918: -> This capability is accommodated in this Standard by using <<per_sorting_sortables-synthetic,synthetic>> sortables.  That is, sortables that are derived or synthesized at the server's discretion.

core/standard/clause_7_building_blocks.adoc
```diff
@@ -909,6 +911,14 @@ sort order.
...
+
+A sever implementing relevance sorting might include, in the list of sortables, a sortable named `relevance` that is dynamically computed for each result in a result set.  The value of the synthetic `relevance` sortable could, for example, be between 0 and 10 with zero indicating least relevance and 10 indicating most relevance.  A `sortBy` clause specified on a request specifying a descending sort be the `relevance` property would present search results from most to least relevant.
```
tomkralidis, line 920: A server implementing relevance ranking might include, in the list of sortables, a sortable named `relevance` that is dynamically computed for each result in a result set.  The value of the synthetic `relevance` sortable could, for example, be between 0 and 10 with zero indicating least relevance and 10 indicating most relevance.  A `sortby` clause specified on a request specifying a descending sort be the `relevance` property would present search results from most to least relevant.

core/examples/json/sortables.json
```diff
@@ -19,6 +19,14 @@
          "description": "This is a synthetic sortable that the server computes on the fly based on the spatial extent of the resource.  Resources without spatial extent have zero area.",
          "type": "number"
       },
+      "relevance": {
+         "title": "Relevance of the result",
+         "description": "This is a synthetic sortable that the server computes on the fly to assign a relevance to the result with respect to the query parameters specified in the original request",
```
tomkralidis, line 24: -> This is a synthetic sortable at the server's discretion that the server dynamically computes to determine relevance to the result with respect to the query parameters specified in the original request

tomkralidis (changes requested):

core/examples/json/sortables.json
pvretano: @tomkralidis are you asking to add "at the server's discretion"?  Why?  All sortables are at the server's discresion.  That is to say that the server decides what is a sortable or not and then create the respond accordingly.

tomkralidis: Yes, however sorting alphabetically or numerically is more predictable than a full text search.  I guess the main point here is that for relevance ranking, the actual algorithm to rank and provide a "score" is, as mentioned today on OAB, "magic".

kalxas: I do not think we need to clarify more.

pvretano: The "magic" is how the value of computed and that is not relevant here just that is computed and exposed as a sortable to allow the results to be ordered accoring to the relevance ranking.  Are you trying to say that the computation is at the discration of the server?  If yes, then I can reword a bit to convey that.

kalxas (approved):

core/standard/clause_7_building_blocks.adoc
kalxas: I am ok to leave as is

core/examples/json/sortables.json
tomkralidis: Works for me, +1

kalxas (approved):

PR #408, opened 2024-12-05
Update instances of "formats" to comply with schema.

### LINKED ISSUE #403: Example "Records Example in GeoJSON" appears to have invalid `formats`
Please let me know if I am putting this issue in the correct place.

In section 8.5.5.4 of the OGCAPI-Records spec, the GeoJSON look like this:

```
  {
  ...
 "properties": {
      ...
      "formats": [
            "CSV",
            "GeoJSON"
          ],
    ...
 }
```

However, the definition of Formats is (via `recordGeoJSON` -> `recordCommonProperties.yaml` -> `format.yaml`):

```
---
type: object
anyOf:
- required:
  - name
- required:
  - mediaType
properties:
  name:
    type: string
  mediaType:
    type: string
```

The example is using a `List<String>`, while the definition shows a `List<Object>`.

I think the example should be more like this (see section 8.2.9):

```
    ...
    "formats": [
        { "name": "TK50" },
        { "name": "GRIB", "mediaType": "application/x-grib" },
        { "mediaType": "application/zip"}
    ]
    ...
```

pvretano:
Closes #403

### DIFF
--- a/core/examples/json/crawlable_catalog/dataset_example/nz-buildings-collection.json
+++ b/core/examples/json/crawlable_catalog/dataset_example/nz-buildings-collection.json
@@ -51,9 +51,17 @@
       ]
     },
     "formats": [
-      "GeoPackage",
-      "GeoJSONSeq",
-      "FlatGeobuf"
+      {
+        "name": "GeoPackage",
+        "mediaType": "application/geopackage+sqlite3"
+      },
+      {
+        "name": "GeoJSONSeq",
+        "mediaType": "application/geo+json"
+      },
+      {
+        "name": "FlatGeobuf"
+      }
     ],
     "license": "CC-BY-4.0"
   },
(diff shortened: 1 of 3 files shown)


kalxas (approved):

tomkralidis (changes requested):
See media type comment/change request.

core/examples/json/crawlable_catalog/dataset_example/nz-buildings-collection.json
```diff
@@ -51,9 +51,17 @@
...
-      "GeoPackage",
-      "GeoJSONSeq",
-      "FlatGeobuf"
+      {
+        "name": "GeoPackage",
+        "mediaType": "application/geopackage+sqlite3"
+      },
+      {
+        "name": "GeoJSONSeq",
+        "mediaType": "application/geo+json"
```
tomkralidis, line 60: Should this be `application/geo+json-seq` as per https://datatracker.ietf.org/doc/html/rfc8142#section-5?

pvretano: Fixed.

tomkralidis (approved):

PR #442, opened 2025-03-01
Fixes of typos

fhoubie:
I've finished reading the spec, great work. 
I've done one change on  ([the Abstract test 30](https://github.com/opengeospatial/ogcapi-records/commit/4fc81436f291379dc80d6e344d216e5a82e00267)) as the bullet list was not correct and it was causing an issue on the numbering. Please review as I'm not a MD expert and it should be better I think. Thank you

### DIFF
--- a/core/standard/abstract_tests/building-blocks/record-collection/ATS_mandatory-properties.adoc
+++ b/core/standard/abstract_tests/building-blocks/record-collection/ATS_mandatory-properties.adoc
@@ -3,7 +3,7 @@
 //[width="90%",cols="2,6a"]
 //|===
 //^|*Abstract Test {counter:ats-id}* |*/conf/record-collection/mandatory-properties-collection*
-//^|Test Purpose |Validate that a catalog objects includes all mandatory properties.
+//^|Test Purpose |Validate that a catalog object includes all mandatory properties.
 //^|Requirement |<<req_record-collection_mandatory-properties-collection,/req/record-collection/mandatory-properties-collection>>
 //^|Test Method |. Construct a path for a catalog object.
 //. Issue a HTTP GET request on that path.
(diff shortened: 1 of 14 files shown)


core/standard/abstract_tests/building-blocks/record-collection/ATS_mandatory-properties.adoc
```diff
@@ -3,7 +3,7 @@
 //[width="90%",cols="2,6a"]
 //|===
 //^|*Abstract Test {counter:ats-id}* |*/conf/record-collection/mandatory-properties-collection*
-//^|Test Purpose |Validate the a catalog objects includes all mandatory properties.
+//^|Test Purpose |Validate that a catalog objects includes all mandatory properties.
```
pvretano, line 6: This should read "Validate that a catalog object includes all mandatory properties."  That is, remove the "s" from the "objects" word.

core/standard/abstract_tests/building-blocks/record-collection/ATS_mandatory-properties.adoc
```diff
@@ -15,7 +15,7 @@
 [%metadata]
 identifier:: /conf/record-collection/mandatory-properties-collection
 target:: /req/record-collection/mandatory-properties-collection
-test-purpose:: Validate the a catalog objects includes all mandatory properties.
+test-purpose:: Validate that a catalog objects includes all mandatory properties.
```
pvretano, line 18: This should read "Validate that a catalog object includes all mandatory properties." That is, remove the "s" from the "objects" word.

fhoubie: Fixed as well

PR #466, opened 2025-04-15
Link schema does not allow CURIEs

joostfarla:
The current OpenAPI schema for links is too restrictive and does not allow CURIE values.
This PR removes the `uri` format, identical to the [schema as published in API Features spec](https://github.com/opengeospatial/ogcapi-features/blob/master/core/openapi/schemas/link.yaml).

### DIFF
--- a/core/openapi/schemas/link.yaml
+++ b/core/openapi/schemas/link.yaml
@@ -8,4 +8,3 @@ allOf:
     properties:
       href:
         type: string
-        format: uri


pvretano:
@joostfarla there is currently a lot of duscussion around CURIEs and not using them ... see [here](https://github.com/opengeospatial/ogcapi-features/issues/975).  I'll need to discuss with the other editors before merging this one.

PR #502, opened 2026-06-02
update stac/records definitions

fmigneault:
## Changes

- Update STAC Example (missing `datetime`, as required by STAC https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#properties)
- Add `/rec/stac-items/conformsto` to include `conformsTo` in STAC Item
- Fix links to registered STAC Extensions and Record conformance classes

## Description

I've added `datetime` to the example since it would not be a valid STAC Item otherwise.

This actually arises another question. Should there be an explicit requirement that the OGC Record including `stac_extensions` should itself be a valid Item/Collection? Technically, there is nothing enforcing this yet, but I don't see why anyone would include a  `stac_extensions` property if the intent isn't for it to be a valid STAC Item/Collection. If anything, allowing to not respect STAC Core would just make record documents more confusing to interpret and will break clients expecting their required fields.

Similarly, there was nothing actually indicating the `conformsTo` URI of OGC API Records (only the server itself had to declare it in `/conformance` endpoint). Since [Records Core](https://docs.ogc.org/is/20-004r1/20-004r1.html#rc_record-core) indicates `conformsTo` is *optional*, I made it a recommendation (`/rec/stac-items/conformsto`).


extensions/stac-extensions/recommendations/stac-items/REC_stac-item-conformsto.adoc
pvretano, line 1: This standard is about allowing OARec records use standard extensions.  As such it should not make and recommendations or requirements about what a STAC item should do.  That is out of scope.

extensions/stac-extensions/requirements/stac-items/REQ_conformance.adoc
pvretano, line 1: Again this should be reworded from the perspective of an OARec record.  Something line "An implementation of a server that supports the ability to generate records that are also valid STAC items SHAL L ...".  Actually this one is my bad I think ... LOL! ;)

extensions/stac-extensions/sections/clause_8_items.adoc
pvretano, line 1: Again, the title of this example should read something like ... "A record with explicit indications of its STAC extensions.".  In other words this is a OARec record that is using STAC extensions.

pvretano (changes requested):
@fmigneault I've made a few comments about the focus of the changes.  They should about **OARec records** that are being extended to either include STAC extensions or to be valid STAC items (so that a STAC client can read them).  Other than that, I think the changes are good!

extensions/stac-extensions/recommendations/stac-items/REC_stac-item-conformsto.adoc
m-mohr: Agreed. But my plan would be to explain the differences in a separate chapter and add some best practices how to use STAC and Records together. Not as requirements though.

fmigneault: I agree. This is why I made it a recommendation. 

I think it would be worth to make this compatibility bidirectional and explicit, since that is realistically the reason why people might be interested in reusing STAC Extensions. 

I can see 2 patterns:

1. A "native" OARec record definition reuses the STAC Extensions. In this sense, it simply borrows the extra properties of said extensions. Because it is primarily an OARec record, it technically does not need a `stac_extensions` field since it can already list the URIs in `conformsTo` and it deals in "OARec record" style interactions with clients. The only reason to employ `stac_extensions` field instead of `conformsTo` is because STAC Items/Collection profile responses are requested/desired.

2. A "native" STAC API definition wants to extend its client compatibility with OARec record without duplicating already available `stac_extensions` in `conformsTo`. In this case, if the `conformsTo` is not extended with OARec record conformance URI, it (or clients using it) cannot be made aware of this support directly. Since `stac_extensions` varies across documents, so should its `conformsTo` (not just at the top-level `/conformance`). 

Therefore, the cleanest and easiest way to extend existing implementations seem to add the missing `conformsTo` and then deal with the contents à la STAC. Adding only a `stac_extensions` field in a pure "OARec record"-style content without respecting all other STAC Core requirements (such that it is a valid STAC Item/Collection) seems rather pointless and misleading.

m-mohr: Yes, indeed. Let's go step by step. The main focus of the extension is (1), let's do that first. And then do (2) etc.

pvretano (approved):
@fmigneault @m-mohr this looks good to me so I am just going to merge it so that you can continue with your work.

